PR #3, opened 2020-04-21
move import path according to repro name

ryanzhang-oss:


.gitignore
```diff
@@ -1,15 +1,25 @@
-/.cache
-/.work
-/_output
-/config/
-/config
-cover.out
-/vendor
-/.vendor-new
```
negz, line 8: I think some or all of the above need to stay in the `.gitignore` as they're generated by the Makefile but shouldn't be checked in.

Makefile
```diff
@@ -92,3 +92,23 @@ oam-runtime.help:
 help-special: oam-runtime.help
 
 .PHONY: oam-runtime.help help-special
+
+generate: controller-gen
+	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
```
negz, line 97: I don't think we need this - the our build submodule make lib calls `go generate`, which calls this per https://github.com/crossplane/oam-kubernetes-runtime/blob/2cccfe365030ec41e27d78ef356f0fc9281816f3/apis/generate.go#L25

ryanzhang-oss: how does this work? I think that the goal is to make this repro independent enough that anyone who is interested in OAM can come to contribute. I am trying to set up a generic system to achieve that and this is just the first step.

negz:
@ryanzhang-oss Can you update the repo name in the `Jenkinsfile` too, please?

PR #5, opened 2020-04-22
Remove c/c dependency and bump crossplane-runtime

hasheddan:
This removes the dependency on c/c such that c/c can import `oam-kubernetes-runtime` as a library. The deleted `modify` and `translate` files will be moved to `addon-oam-kubernetes-remote` as they must import types from c/c itself (i.e. `KubernetesApplication`).


apis/generate.go
```diff
@@ -20,20 +20,13 @@ limitations under the License.
 // https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
 
 // Remove existing CRDs
-//go:generate rm -rf ../cluster/charts/crossplane-types/crds
+//go:generate rm -rf ../cluster/charts/oam-kubernetes-runtrime/crds
```
muvaf, line 23: typo

hasheddan: good catch, thanks! fixed :+1:

PR #6, opened 2020-04-23
Remove generated CRDs

hasheddan:
As a library, oam-kubernetes-runtime should not generate its own CRDs. Consumers of the library should generate and install the CRDs from the types defined in the library. See https://github.com/crossplane/crossplane/pull/1436 for more information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ vendor/
 .DS_Store
 bin/
 tmp/
+crds/
 
 # Vscode files
 .vscode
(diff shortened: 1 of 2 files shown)


ryanzhang-oss:
Does "make generate" still work after this?

hasheddan:
@ryanzhang-oss yes, it will still generate methodsets, just not the CRDs 👍

ryanzhang-oss:
> @ryanzhang-oss yes, it will still generate methodsets, just not the CRDs 👍

How do you envision the library users to get the CRD then?

negz:
> How do you envision the library users to get the CRD then?

It would be up to the actual runtime to generate and package the CRDs; see https://github.com/crossplane/crossplane/pull/1436 for an example of how that can be done. A library doesn't seem like the right place for runtime configuration to live.

PR #7, opened 2020-04-23
Add readme and an example

ryanzhang-oss:
Add readme and an example


README.md
```diff
@@ -1,2 +1,74 @@
-# oam-runtime
-A set of libraries for building OAM runtimes
+# oam-kubernetes-runtime Project
+
+The OAM Kubernetes runtime project is a collection of Golang help libraries and official api type
```
hongchaodeng, line 3:
```suggestion
The OAM Kubernetes runtime project is a collection of Golang helper libraries and api
```

README.md
```diff
@@ -1,2 +1,74 @@
...
+ definitions for building an OAM Kubernetes runtime that is compatible with the latest stable
```
hongchaodeng, line 4:
```suggestion
 definitions for building an OAM Kubernetes runtime
```

README.md
```diff
@@ -1,2 +1,74 @@
...
+  OAM spec.
```
hongchaodeng, line 5:
```suggestion
```

README.md
```diff
@@ -1,2 +1,74 @@
...
+# oam-kubernetes-runtime Project
+
+The OAM Kubernetes runtime project is a collection of Golang help libraries and official api type
+ definitions for building an OAM Kubernetes runtime that is compatible with the latest stable
+  OAM spec. 
+
+## Prerequisites
+
+1. go version 1.12+
+2. Kubernetes version v1.15+ with KubeCtl configured
```
hongchaodeng, line 10:
```suggestion
2. Kubernetes version v1.15+ with kubectl configured 
```

pkg/examples/containerized-workload/sample_application_config.yaml
```diff
@@ -0,0 +1,66 @@
+---
```
resouer, line 1: Please split definition yamls from app_config.yaml, they do not belong to same world.

pkg/examples/containerized-workload/main.go
```diff
@@ -0,0 +1,60 @@
+package main
```
resouer, line 1: examples should not live in `pkg` in general. I'd suggest:

```
developing/sample_controller/main.go
```

README.md
```diff
@@ -1,2 +1,74 @@
-# oam-runtime
-A set of libraries for building OAM runtimes
+# oam-kubernetes-runtime Project
```
resouer, line 1:
```suggestion
# OAM Kubernetes Runtime
```

README.md
resouer, line 3: My suggestion:

The OAM Kubernetes Runtime is the building block of data model in [Crossplane](https://github.com/crossplane/crossplane). It's basically a collection of Golang help libraries and api types to make implementing [Open Application Model](https://github.com/oam-dev/spec) easier on Kubernetes. 

OAM Kubernetes Runtime is designed to be platform builder facing and embedded into a larger system, for end-users facing experience of OAM, please always check [Crossplane official documentation](https://crossplane.github.io/docs).

README.md
```diff
@@ -1,2 +1,74 @@
...
+```
+
+And a manualscalertraits looking like below
+```
+kubectl get manualscalertraits.core.oam.dev
+NAME                      AGE
+example-appconfig-trait   54s
+```
+
+## Goals
```
resouer, line 64: Let's delete this section, I've add a suggestion to include goals at the beginning.

README.md
```diff
@@ -1,2 +1,74 @@
...
-A set of libraries for building OAM runtimes
+# oam-kubernetes-runtime Project
+
+The OAM Kubernetes runtime project is a collection of Golang help libraries and official api type
+ definitions for building an OAM Kubernetes runtime that is compatible with the latest stable
+  OAM spec. 
+
+## Prerequisites
+
+1. go version 1.12+
```
hasheddan, line 9: I believe we are using go `1.13` here, but I don't think this a strict prerequisite for consuming?

pkg/examples/containerized-workload/main.go
hasheddan: I would advocate for naming the top-level directory `examples/` if it is all the same to everyone. Another option would be to just point to Crossplane's usage of it as an example, then provide smaller pointers in documentation

README.md
```diff
@@ -1,2 +1,74 @@
...
+
+make
+```
+
+* Install CRDs to your Kubernetes cluster
+
+```shell
+kubectl apply -f cluster/charts/oam-kubernetes-runtime/crds/
+
+```
```
hasheddan, line 34: These are being removed as it makes more sense for an implementation to generate their own CRDs

ryanzhang-oss: in this case, I will add a make CRD command in the make file so that this example can be self-contained.

README.md
ryanzhang-oss: I believe this is related to the go module usage.  Maybe go 1.11 also works but I haven't tried.

Makefile
```diff
@@ -92,3 +92,31 @@ oam-runtime.help:
...
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
+endif
+
+# find or download controller-gen
+# download controller-gen if necessary
+controller-gen:
```
hasheddan, line 105: I like your pattern of just adding `/crds` to the `.gitignore`, I can re-enable the generation here and we can just ignore the output: https://github.com/crossplane/oam-kubernetes-runtime/pull/6/files#diff-26d5f9a9604b23990843fe2c336219caR23

We already pin `controller-gen` in `generate.go` so no need to add this portion to the `Makefile`.

README.md
```diff
@@ -1,2 +1,75 @@
-# oam-runtime
-A set of libraries for building OAM runtimes
+# OAM Kubernetes Project
```
resouer, line 1: It's not common to say xxx project in general, especially for a building block library.

```suggestion
# OAM Kubernetes Runtime
```

Makefile
```diff
@@ -91,4 +91,4 @@ oam-runtime.help:
 
 help-special: oam-runtime.help
 
-.PHONY: oam-runtime.help help-special
+.PHONY: oam-runtime.help help-special
```
resouer, line 94: nits: still have several no blank line issues here and below

README.md
```diff
@@ -1,2 +1,75 @@
...
+
+The OAM Kubernetes runtime project is a collection of Golang helper libraries and OAM api type
+ definitions. 
+
+It is designed to help OAM platform builders rather than being used directly by developers
```
resouer, line 6: These two paragraphs are verbose and repeating. 

Could you compress the main points in one paragraph (i.e. this repo is platform builder facing and important for ensuring interoperability in OAM platforms)?

.gitignore
```diff
@@ -15,6 +15,7 @@
 cover.out
 /vendor
 /.vendor-new
+/crds
```
hasheddan, line 18: /crds is included below now so this can be removed :)

README.md
ryanzhang-oss: I actually expended a bit here to make our point loud and clear.

resouer: That should be the reason I feel it's verbose :-) Personally I'd prefer one sentence/paragraph to tell out the main points.

README.md
```diff
@@ -1,2 +1,75 @@
-# oam-runtime
-A set of libraries for building OAM runtimes
+# OAM Kubernetes Runtime
+
+The OAM Kubernetes runtime project is a collection of Golang helper libraries and OAM api type
+ definitions.
```
hasheddan, line 4:
```suggestion
definitions. 
```

README.md
```diff
@@ -1,2 +1,75 @@
...
+
+* Apply the sample application configurations
+
+```
+kubectl apply -f examples/containerized-workload/ 
+```
+
+* Verify that corresponding CRs are emitted. 
+
+You should see a containerizedworkloads looking like below
```
hasheddan, line 55:
```suggestion
You should see a `ContainerizedWorkload` looking like below
```

README.md
```diff
@@ -1,2 +1,75 @@
...
+```
+kubectl get containerizedworkloads.core.oam.dev  
+NAME                         AGE
+example-appconfig-workload   12s
+```
+
+And a manualscalertraits looking like below
```
hasheddan, line 62:
```suggestion
And a `ManualScalerTrait` looking like below
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
```
negz, line 31: Why are we installing dep here. This project uses Go modules.

README.md
negz: What is the difference between an "OAM platform builder" and a "developer"? Does developer mean "folks developing OAM", or is it synonymous with "end-user" in this context?

README.md
```diff
@@ -1,2 +1,74 @@
...
+
+* Fork and Clone this project
+
+* Build the library 
+
+```shell
+make submodules 
+
+make
+```
```
negz, line 32: `make test` and `make reviewable` are also handy targets that may be worth documenting. The former runs the unit tests, while the latter runs linters, generation, etc to catch some of the things that might cause CI to fail.

README.md
```diff
@@ -1,2 +1,74 @@
-# oam-runtime
-A set of libraries for building OAM runtimes
+# OAM Kubernetes Runtime
```
negz, line 1: A Github badge just a link to https://pkg.go.dev/mod/github.com/crossplane/oam-kubernetes-runtime may help folks discover the API docs.

.github/workflows/go.yml
hasheddan: Looks like this just a generic github workflow that will only install dep if a Gopkg.toml is present

README.md
ryanzhang-oss: a developer is an OAM application developer, aka "end-user".  Do you prefer to the term "application developer" or "end-user"?

.github/workflows/go.yml
ryanzhang-oss: Github auto-generated this, not sure if it's safe to delete.

README.md
ryanzhang-oss: sure, we can add some instructions in the "how to contribute" section.

README.md
ryanzhang-oss: what is a Github badge?

PR #8, opened 2020-04-25
Add e2e test framework and use github action CI

ryanzhang-oss:


e2e-test/example_test.go
```diff
@@ -0,0 +1,229 @@
+package e2e_test
```
hongchaodeng, line 1: example_test.go as file name is confusing...
Maybe just e2e_test.go or appconfig_test.go?

ryanzhang-oss:
> @ryanzhang-oss thanks for working on this! How committed are you to `ginkgo` / `gomega`? I know they are used prevalently throughout the k8s ecosystem, but we have done a bit of work on doing integration tests in a table-based format (ref: https://github.com/crossplane/crossplane-runtime/tree/master/pkg/test/integration // https://github.com/crossplane/provider-gcp/tree/master/test/integration). Is this something you would be interested in trying out? I personally feel that is is much easier to troubleshoot errors with it, but like I said lots of folks do like `ginkgo`. I would be happy to open a PR with some example tests here for comparison if you like :)

Sure, feel free to open a PR with examples. Personally, I have no preference as I just learnt `ginkgo` like last week. However, as you mentioned, it seems to be the most prevalent one in the community. I think (in theory) adopting it should help (or at least not a barrier) attract community members to contribute here. Just my 2 cents.

e2e-test/e2e_test_suite_test.go
```diff
@@ -0,0 +1,86 @@
+package e2e_test
```
draveness, line 1: nit: `_` as a package name is not encouraged in golang, e2etest could be a better substitute.

FYI: https://blog.golang.org/package-names

wonderflow:
> @ryanzhang-oss thanks for working on this! How committed are you to `ginkgo` / `gomega`? I know they are used prevalently throughout the k8s ecosystem, but we have done a bit of work on doing integration tests in a table-based format (ref: https://github.com/crossplane/crossplane-runtime/tree/master/pkg/test/integration // https://github.com/crossplane/provider-gcp/tree/master/test/integration). Is this something you would be interested in trying out? I personally feel that is is much easier to troubleshoot errors with it, but like I said lots of folks do like `ginkgo`. I would be happy to open a PR with some example tests here for comparison if you like :)

In k8s community, it's common to have different test framework in one project, we don't need to have big opinion about test. I'm afraid blocking PR contribution  with a reason about test framework will discourage community guys to contribute.

However, I don't have any opinions about test framework, but test coverage is really important. Do you have any other concerns about it? @hasheddan

ryanzhang-oss:
merge https://github.com/crossplane/oam-kubernetes-runtime/pull/17 instead

PR #20, opened 2020-05-13
add child resource kind to the workload definition

ryanzhang-oss:
add child resource kind to the workload definition according to  https://github.com/crossplane/crossplane/pull/1454


wonderflow:
@hasheddan  Let's comment more on this PR, then `dev` branch can move on quickly.

hongchaodeng:
lgtm

PR #21, opened 2020-05-16
Add ServiceInjector TranslationWrapper

hasheddan:
The ServiceInjector translation wrapper was not carried over from crossplane/crossplane when oam-kubernetes-runtime was bootstrapped. This adds it. See previous implementation: https://github.com/crossplane/crossplane/blob/8f7ffc42936f8ae96612e82bae5e1437065bdbe5/pkg/oam/workload/translate.go


hongchaodeng:
lgtm

PR #24, opened 2020-05-18
design: resource dependency in OAM

hongchaodeng:
Coming from https://github.com/oam-dev/spec/pull/326.
Similar to https://github.com/crossplane/oam-kubernetes-runtime/pull/23.

Since it's no change on OAM spec and more implementation focused, we are moving the design into here.


doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+
+The functionality in OAM runtime is an internal implementation.
+Either DataOutput trait or a potential field `outputs` are representation layer.
+We could map either into underlying DataOutput resources.
+
+It's inevitable that an intermediate phase where both exists could happen.
+But they don't both exist in the same ApplicationConfiguration.
+Only one could exist and we don't need to handle any conflicts.
+
+
```
wonderflow, line 300:
```suggestion
Once that upgrade is needed, we could add an webhook to do this translation. So the DataOutput trait could be translated to native OAM fields.
```

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+    kind: MyWorkload
+    metadata:
+      name: my-web
+      annotations:
+        "core.oam.dev/datainputs": "[{\"toFieldPaths\": [\"spec.connSecret\"],\"fromDataOutput\": \"mysql-conn-secret\"}]"
+    ```
+
+    Note that my-web won't be created since it has a DataInputs dependency to satisfy.
+
+2. Add a new CRD `DataOutput` with the following definition:
```
wonderflow, line 137: I think this CRD need to be named as `DataInput`. Because it is bind to components who need to consume these data.

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+        metadata:
+          name: my-rds
+    ---
+    # In ApplicationConfiguration
+    components:
+    - componentName: my-web
+      traits:
+      - trait:
+          apiVersion: experimental.core.oam.dev/v1alpha2
+          kind: DataOutput
```
wonderflow, line 255: A component binding with a `DataOutput` trait seems tell others this component will output something. But in fact the component is consuming data from this trait. So I think naming is with `DataInput` is more intuitive.

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+This is similar to [kubernetes#65502](https://github.com/kubernetes/kubernetes/issues/65502) , the issue describes dependencies between containers on the same Pod, while we are discussing dependencies between components.
+
+
+## Proposal
+
+The overall idea is to have each resource object specify data inputs coming from other resources' fields.
+The OAM runtime will build the dependency graph accordingly and only create resources once the data inputs are ready.
+To achieve this, we propose:
+
+1. Add the following `DataInputs` into each resource.
```
hasheddan, line 97: It is not clear to me what "each resource" refers to here. It seems like this would potentially be surfaced on a `Component` similar to how `parameters` is currently, but it is not present on the `Component` examples below.

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+      name: my-web
+      ..
+    spec:
+      connSecret: mysql-conn # patched by OAM runtime
+    ```
+
+3. Below we show how to combine the above concepts (DataInput, DataOutput) into OAM.
+
+    In OAM Component we put a syntax-sugar on existing `Parameters` fields:
+    as long as the name of the Parameter matches the name of the DataOutput, it becomes a DataInput.
```
hasheddan, line 224: This seems to answer my question above, but can you also explicitly provide `dataInputs`?

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+    metadata:
+      name: my-web
+    spec:
+      workload:
+        kind: MyWorkload
+        metadata:
+          name: my-web
+      parameters:
+      - name: mysql-conn-secret # this matches the name of DataOutput
+        fieldPaths: ["spec.connSecret"]
```
hasheddan, line 238: If a user does not provide a `DataOutput` trait, could they provide a parameter value here to set this explicitly?

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+    apiVersion: experimental.core.oam.dev/v1alpha2
+    kind: DataOutput
+    metadata:
+      name: mysql-conn-secret
+    spec:
+      resource:
+        apiVersion: alibaba.oam.crossplane.io/v1alpha1
+        kind: RDSInstance
+        name: my-rds
+      fieldPath: "status.connSecret"
```
hasheddan, line 183:
```suggestion
      fieldPath: "spec.writeConnectionSecretToRef"
```
https://doc.crds.dev/github.com/crossplane/provider-alibaba/config/crd/database.alibaba.crossplane.io_rdsinstances.yaml#spec-writeConnectionSecretToRef

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+          apiVersion: experimental.core.oam.dev/v1alpha2
+          kind: DataOutput
+          metadata:
+            name: mysql-conn-secret # this matches the name of a component's parameter
+          spec:
+            resource:
+              apiVersion: alibaba.oam.crossplane.io/v1alpha1
+              kind: RDSInstance
+              name: my-rds
+            fieldPath: "status.connSecret"
```
hasheddan, line 263:
```suggestion
            fieldPath: "spec.writeConnectionSecretToRef"
```

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+    metadata:
+      name: my-rds
+    spec:
+      workload:
+        kind: RDSInstance
+        metadata:
+          name: my-rds
+    ---
+    # In ApplicationConfiguration
+    components:
```
hasheddan, line 250: In this case I am not sure how much using the `DataOutput` is helping the end user. They still have to be aware of the connection secret field of the `RDSInstance` which is a level lower than the `Component` they are using. It seems like a better experience would be for the `Component` author to surface a `secretName` parameter for the `my-rds` and `my-web` `Components`, then the end user could just provide the same value for each. The user would only be required to know about the components rather than knowing about the underlying resource types, which seems like the abstraction that we want at the `ApplicationConfiguration` level.

hasheddan: Consider the experience here vs this from the quick start with service tracker:

```yaml
spec:
  components:
    - componentName: data-api-database
      parameterValues:
        - name: database-secret
          value: tracker-database-secret
    - componentName: data-api
      parameterValues:
        - name: database-secret
          value: tracker-database-secret
```

doc/design/resource-dependency.md
hongchaodeng: The name DataInput is already taken in the input resource side. To correspond to that, this is named to DataOutput. I don't think we should change that.

doc/design/resource-dependency.md
hongchaodeng: same as above. Gonna resolve this and merge conversation.

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,343 @@
...
+It's inevitable that an intermediate phase where both exists could happen.
+But they don't both exist in the same ApplicationConfiguration.
+Only one could exist and we don't need to handle any conflicts.
+
+
+## Solution examples
+
+In this section we are providing more solution examples for dependency use cases.
+
+1. Service binding trait:
```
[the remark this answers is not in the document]
hongchaodeng: > Maybe we should prioritize adding initial support for it here?

This is just an example use case. The dependency feature is not tightly bound to it.

doc/design/resource-dependency.md
hongchaodeng: Both Components and Traits are k8s resources.
It just means the resources that needs data input.

doc/design/resource-dependency.md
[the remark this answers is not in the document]
hasheddan: Makes sense :) It just makes it a little hard to reason about when the example is not currently supported, especially for folks that are not familiar with `trait-injector`. I get the idea though :+1:

doc/design/resource-dependency.md
hongchaodeng: Yes. This is also put into the annotations.

Actually parameter is very Component specific. What's designed here is a generic resource dependency engine that as long as some resource has DataInputs it will build corresponding dependency.

doc/design/resource-dependency.md
hongchaodeng: Currently parameterValue does not support fields like this.
We model it as a trait to avoid changing the spec.

doc/design/resource-dependency.md
hongchaodeng: Well, this is an example to show field passing.
We have many use cases that passes from status, which means they are not available at the beginning.

doc/design/resource-dependency.md
hongchaodeng: I totally understand. But this is something we could not enforce users to adopt. For example, in Alibaba cloud, the ID of RDS would not be available at the beginning until RDS instance is created. This has to be passed from the status.

doc/design/resource-dependency.md
hongchaodeng: TODO: I will update the doc.

doc/design/resource-dependency.md
hongchaodeng: TODO: will update the doc

doc/design/resource-dependency.md
hasheddan: Thanks :)

doc/design/resource-dependency.md
hongchaodeng: Add a section to discuss generic engine design.

hongchaodeng:
Updated to address comments and improve sentences.

doc/design/resource-dependency.md
```diff
@@ -0,0 +1,350 @@
...
+
+      Kind string `json:"kind"`
+
+      // Name indicates the name of the object in the same namespace.
+      Name string `json:"name,omitempty"`
+    }
+
+    type DataOutputStatus struct {
+      // Ready indicates whether the field data of the source resource is ready.
+      Ready bool `json:"ready"`
```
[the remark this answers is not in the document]
hongchaodeng: Agree.
I would change this to standard k8s Conditions.

hongchaodeng:
Hi @negz @hasheddan  . Thanks for the comments. Let me answer them all here:

>  Is the below understanding correct?
> 1. We'll overload the existing concept of component parameters to also reference a DataOutput
> 2. The AppConfig controller will discover params that are actually data inputs (i.e. references to a DataOutput).
> 3. The AppConfig controller will will wait until the referenced DataOutput is ready, then inject the value from the DataOutput into the component's workload template before rendering the workload.

Correct.

> What controller will be responsible for writing DataOutput CRs?

The OAM user will write it.
Let me give an example. In a PaaS platform like Elastic Beanstalk, it will ship the web application and RDS database together. The PaaS platform then prepares an ApplicationConfiguration yaml file with components (incl. Parameters) and traits (incl. DataOutput).

> Why do we need to annotate workloads with this information after they're created?

Parameter is Component specific. Nonetheless, we also have use cases for Component2trait, trait2trait, and Parameter doesn't work for them. In fact, their underlying implementation are the same.

As a result, the generic resource dependency engine would make most sense here. By separating the core implementation from the representation layer, we could reuse the core service and have different API layers.

The core engine includes the dependency graph, DataInputs, DataOutputs. The representation layer includes Component parameters and other similar APIs.

> The implicit param to data output relationship feels potentially surprising to users. Could we make it explicit with a small, backward compatible spec change?

The main point is not to change spec at all to move forward and validate the idea first. In the community meeting that we discussed and agreed to have implementation first and validate the ideas before making any spec changes.

> You state that this (DataInputs) needs to be marshalled and put into an annotation key. Could you explain (ideally in the document) why? What controller would read this annotation after it is written?

This is a compromise for initial implementation. For experimental features we will all do it in annotations in oam-kubernetes-runtime. Once validates, these schema will be added to the spec.

> I think what @hasheddan is asking is whether this parameter acts like a "normal" parameter if there is no corresponding DataOutput. i.e. Will the user be able to supply a parameter value for mysql-conn-secret explicitly when they write their ApplicationConfiguration?

Yeah of course.
However, there is another background information that we are actually dropping use cases for parameterValues. See [this](https://github.com/oam-dev/spec/issues/336#issuecomment-608802349). We would model this as traits.

> I'm having a hard time understanding what CR(s) you're proposing will support this datainputs field. Does it ever exist in the YAML format you show here, or is it only ever persisted as an annotation?

It is persisted as an annotation.
Put it as yaml for better reading.

hongchaodeng:
Hi @negz. Thanks for the reply.

I would prefer NOT to make DataInput as another CR which would involve another resource ref as below:

```yaml
kind: DataInput
spec:
  resourceRef: # resource that needs input
  fromDataOutput: ...
```

I do agree with you on the ParameterValue solution. Could we make it as a step-by-step progress? For example, in alpha we make it as annotations to validate the idea. In beta after like two or three months, we make it into the spec in ParameterValue. How does it sound?

The major blocker is that the spec is just released as v1alpha2. It might take another cycle and we don't want to make major changes to it for now.

negz:
> For example, in alpha we make it as annotations to validate the idea. In beta after like two or three months, we make it into the spec in ParameterValue. How does it sound?

Should we discuss this at the next community meeting - or setup a call later next week (after Weds)? I apologise for missing the discussion that happened at the last one! I agree that we shouldn't block on updating the spec, but I think we can find a better solution than using annotations, and I don't want to set a precedent if we agree that we don't really like that solution long term.

hongchaodeng:
Sure. Will set up a meeting next week to involve @vturecek and @artursouza too

artursouza:
The latest version for this proposal (using new CRDs for DataOutput) is super confusing. As an app developer I would simply use another framework if I was comparing solutions to model my app on K8s. Just my opinion here.

On practical terms: if we need all this workaround to make this work without changing the spec, we should probably step back and change the spec to have a first-class experience. Also, the way we reference params via fieldPaths is really confusing, even more now - where we mix how things are referenced (from consumer to resource vs from resource to consumers).

wonderflow:
@artursouza  Totally agree if we could change spec and make it as experimental feature. For component params(parameters), we are planning to remove it. As it not only confusing on fieldPath, but also confusing on [component versioning mechanism](https://github.com/crossplane/oam-kubernetes-runtime/pull/29#issuecomment-634454644).

hongchaodeng:
After discussion with @artursouza , we will also change the DataOutput to be (experimental) fields.

resouer:
A offline discussion, how about we add experimental field in OAM spec like this:

| Attribute | Type | Required | Default Value | Description |
|-----------|------|----------|---------------|-------------|
| `workload` | [`Workload`](#workload) | Y | | Declaration of settings that should be passed to the workload runtime|
| `parameters` | [`[]Parameter`](#parameter) | N | | The component's configuration options. |
|`policies` | [`[]Policy`](#policies) | N | | <kbd>EXPERIMENTAL</kbd>  The operational policies for this component. |

hongchaodeng:
> how about we add experimental field in OAM spec like this:

Sounds great!

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+        - trait:
+            apiVersion: example.io/v1alpha1
+            kind: EtcdBackup
+            metadata:
+              name: my-etcd-backup
+          parameters:
+          - valueFrom: # An EtcdBackup trait has dependency on some DataOutput to fill its fields.
+              dataOutputName: etcd-conn-secret
+            toFieldPaths:
+            - spec.connSecret
```
wonderflow, line 236: if trait can have `parameters` to put in data, then trait should also be able to have `dataoutputs` to output data.
For example, RDS as a trait.

wonderflow: Oh, I see the ServiceBinding example, really cool! Ignore my comment here.

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+    Component:
+
+    ```yaml
+    kind: Component
+    metadata:
+      name: my-web
+    spec:
+      parameters:
+      - name: my-secret-binding
+        fieldPaths: [] # This parameter is for ordering purpose only
```
ryanzhang-oss, line 347: this seems a bit hacky as there is no data passing

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+            spec:
+              from:
+                secret:
+                  name: my-secret
+              to:
+                env: true
+          dataOutputs:
+            - name: my-secret-binding
+              fieldPath: status.ready
+    ```
```
ryanzhang-oss, line 336: why is there no `parameterValues` field that contains `valueFrom`  in the component here? Is it just implied?

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+        - name: etcd-conn-secret
+          fieldpath: "status.connSecret"
+
+        traits:
+        - trait:
+            apiVersion: example.io/v1alpha1
+            kind: EtcdBackup
+            metadata:
+              name: my-etcd-backup
+          parameters:
```
ryanzhang-oss, line 232: why is this a `parameter` instead of `parameterValue`? Also, I don't seem to see a new `parameter` field in the `traitDefinition` object

design/resource-dependency.md
ryanzhang-oss: synced with @hongchaodeng offline.  I get that this is a work-around for pure dependency cases. My take is that since OAM adopts the declarative model, one doesn't really have to enforce this type of dependency as it will play out at the end. The controllers should just give up reconcile for a few times before they eventually succeed

design/resource-dependency.md
ryanzhang-oss: I get that this is pretty much a `syntax sugar` which basically avoids writing another `parameterValues` in the component. I am not sure if I like it or not. On one hand, it saves some repetitive work but on the other hand, it is inconsistent with the way parameters are assigned in components.

design/resource-dependency.md
ryanzhang-oss: I think this should be called `parameterValues` so that the keyword `parameter` only means "declare a parameter" in OAM. IMO, it is too confusing for `parameter` to mean `declare` (in Component) or `assign` (in Trait) depends on which resources it appears in the OAM spec.

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+    
+    ```yaml
+    kind: MyWorkload
+    metadata:
+      name: my-web
+    spec:
+      connSecret: mysql-conn # patched by OAM runtime
+    ```
+
+3. Similary, add fields `dataOutputs` and `parameters` in Trait configuration to specify the dependency on data output objects.
```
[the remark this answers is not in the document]
ryanzhang-oss: I get that since `trait` is supposedly maintained by `application operator` only, we don't need to have a `declare` phase and every field in a `trait` is parameterizable. I am not entirely convinced this is true though.

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+      - name: database-secret # workload's fieldPaths needs to be filled
+        fieldPaths:
+        - spec.connSecret
+    ---
+
+    kind: ApplicationConfiguration
+    spec:
+      components:
+      - componentName: my-web
+        parameterValues:
```
resouer, line 165: To echo @ryanzhang-oss 's [comment](https://github.com/crossplane/oam-kubernetes-runtime/pull/24/files#r432096300), I'd propose we put data input/output model alongside with `parameters/parameterValues`, then we will gradually deprecate the later feature in the upcoming releases. In that case, we will have a consistent model for both component and trait in dependency description. My assumption is developer doesn't need to declare `parameters` in component to make data input/output model work. 

But what I'm not sure is: who should declare data input/output? In this proposal, it's operator/system. I have feeling that maybe it should be developer? In that case, we should define a `Application` or `ComponentGroup` first and move data input/output. Let's discuss this part.

Note that, I still think it's a valid case for developer to claim `parameters` as "overridable", but  it's very tricky to guarantee this in implementation (AppConfig controller). I'd suggest we merge this use case into "policies" and enforce it by policy engine (if possible).

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+
+An etcd cluster would be created via EtcdCluster CR, and automatic periodic etcd backup would be made via EtcdBackup CR.
+In OAM, we model EtcdCluster as Component and EtcdBackup as Trait.
+However, the EtcdBackup depends on the connection secret specified in `status.connSecret` field of EtcdCluster.
+In other words, the EtcdBackup trait depends on EtcdCluster component.
+
+## Proposal
+
+The overall idea is to have each resource specify data inputs coming from data outputs (i.e. other resources' fields).
+The OAM runtime will build the dependency graph accordingly and only create dependent resources once the data outputs are ready.
```
zzxwill, line 103: `dependency graph` means DAG(directed acyclic graph)?

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+
+
+## Use Cases
+
+### 1. Web app and database
+
+An applicationConfiguration consists of two components: web and db (database).
+Web component should not be created until db component is created,
+including public endpoint and connection secret to be made ready.
+
```
rynowak, line 54: We should also consider traditional REST APIs (need the URIs of my other services) as a keystone scenario for any depenedency use case. It's exceedingly common (we already have a sample) and it seems like we could avoid the incredible level of coupling and boilerplate it requires today: https://github.com/oam-dev/samples/blob/master/2.ServiceTracker_App/ApplicationConfiguration/tracker-app-config.yaml#L34

The proposal here seems like it could already help with this, but I want to make sure we're considering it as a headline scenario.

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+They have strict ordering:
+
+```
+preprocess -> training -> serving
+```
+
+Each component has to wait for data to be finished processing by the previous stage.
+Today, ML frameworks including Tensorflow do not handle such dependency issue.
+They rely on external tools (e.g. Argo Workflow) to handle it.
+If we want to model them as Components in the same ApplicationConfiguration, the external tools could not help in this case.
```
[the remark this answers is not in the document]
rynowak: I think there are some good examples here already (4 & 5) that describe scenarios where ordering matters. The workflow example seems like it's better solved by making it possible to just use a workflow engine (possibly as an extended OAM workload type).

design/resource-dependency.md
rynowak: > But what I'm not sure is: who should declare data input/output? In this proposal, it's operator/system. I have feeling that maybe it should be developer? In that case, we should define a Application or ComponentGroup first and move data input/output. Let's discuss this part.

100% agree this needs more thought. 

I'd make the case that it's the developer - developers already think of applications this way and already solve these problems.  

Ask the question - if I'm the developer - *how do I run my application with dependencies today to test a change?* They are likely writing a mock, using docker-compose, or just launching multiple containers/processes in on their own machine.  For a database, they are likely running one in a container or against a staging instance in the cloud.

You can't get very far as a developer without knowing the *identity* of the other services your component talks to. Most often (based on my customer conversations) the components they communicate with are also written by them/their-team.

It's a big win if we can keep these concepts high level where possible - *"I need to talk to weather-api"* allows us to build a better system than *"I need to declare a parameter to set an environment variable so that someone else with a different job can satisfy a binding from another component..."*.

resouer: We had a offline but similar discussion around the persona here. So one reasonable approach is:

```yaml
kind: Component
spec:
  workload:
      ...
  dataOutputs:
  - name: etcd-conn-secret
  ...
```

```yaml
spec:
  components:
  - componentName: my-etcd-cluster
    traits:
    - trait:
        apiVersion: example.io/v1alpha1
        kind: EtcdBackup
        metadata:
          name: my-etcd-backup
          ...
      dataInputs:
      - from: etcd-conn-secret
        toFieldPaths:
        - spec.connSecret
```

i.e. `Component` and `Trait (AppConfig)` has its own data inputs/outputs claim individually. /cc @hongchaodeng

hongchaodeng: Shouldn't Component's data outputs also sit in ApplicatoinConfiguration:

```
spec:
  components:
  - componentName: my-etcd-cluster
    dataOutputs:
      - name: etcd-conn-secret
        ...
    traits:
    - trait:
        apiVersion: example.io/v1alpha1
        kind: EtcdBackup
        metadata:
          name: my-etcd-backup
          ...
      dataInputs:
      - from: etcd-conn-secret
        toFieldPaths:
        - spec.connSecret
```

resouer: I assume @rynowak 's point is component dependency is knowledge of developers. So the components' data inputs/outputs should be part of `kind:component` object.

design/resource-dependency.md
```diff
@@ -0,0 +1,348 @@
...
+
+1. **Describe ordering for k8s-style resources**.
+  The ordering between resources means creating one resource after another.
+  For example, resource A could not be created until resource B has been created and provided outputs.
+
+2. **Describe field data passing between resources**.
+  It is still not enough to solve only ordering problem. We also need to handle field data passing.
+  For example, resource A's field "spec.connSecret" relies on resource B's field "status.connSecret". 
+  This means the field data of one resource is passed from another field of other resource and needs to wait the field data to be ready.
+  In fact, ordering in the first goal can be modeled as data dependency as well except that the receiver doesn't use the data.
```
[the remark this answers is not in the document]
hongchaodeng: I totally understand your point, and this issue has been discussed in last few community meetings.

The reason is that we can't enforce all applications to be fault tolerant and keep retrying. There are many scenarios discussed below that needs such ordering capability. But they can tolerate if the service has a later outage which is infrequent.

design/resource-dependency.md
hongchaodeng: That's the purpose of this proposal.

design/resource-dependency.md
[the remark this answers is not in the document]
hongchaodeng: This proposal gives a representation/model to do the work in OAM.
Of course, the underlying runtime can implement it using external workflow engine.

design/resource-dependency.md
hongchaodeng: yes.

design/resource-dependency.md
hongchaodeng: We can definitely add data inputs in Component definition.
But in this proposal, we would solve the Component configuration part first.
For the use cases described here, it is the operator role that knows the relationship between Components and traits.

design/resource-dependency.md
hongchaodeng: Going to change this to data input fields

hongchaodeng:
Update the proposal to use native fields `dataInputs` and `dataOutputs` after syncing offline.
We will discuss this in tomorrow's community meeting again.

wonderflow:
I think this proposal is almost good, but we'd better give some convert ability in Service Binding trait.


For example, convert key/value in K8s secret to status. And let datainput/dataoutput to pass through.

```
    spec:
      components:
      - componentName: my-web
        dataInputs:
        - fieldPaths: [] # This data input is for ordering purpose only
          valueFrom:
            dataOutputName: my-secret-key
        traits:
        - trait:
            apiVersion: core.oam.dev/v1alpha1
            kind: ServiceBinding
            metadata:
              name: my-secret-binding
            spec:
              from:
                secret:
                  name: my-secret
              to:
                status: true                
          dataOutputs:
            - name: my-secret-key
              fieldPath: status.map.key
```

hongchaodeng:
Merging this proposal as it gets majority consensus (Alibaba+Microsoft).
I have tried to ping more folks for review but haven't received any further feedback yet.
We are going to do implementation as it is.
Any further feedback is welcome and I will adapt implementation to it.

PR #29, opened 2020-05-22
design: one pager component mutable and versioning mechanism for OAM

wonderflow:

Coming from ["OAM component versioning mechanism"](https://github.com/oam-dev/spec/issues/336) and ["The Component should mutable"](https://github.com/oam-dev/spec/issues/350).

Since we discussed a lot and have a conclusion with OAM spec changes done. I write this proposal to make the implementation details more clear.

### DIFF
--- a/design/one-pager-component-mutable-and-versioning.md
+++ b/design/one-pager-component-mutable-and-versioning.md
@@ -1,0 +1,497 @@
+# Component mutable and versioning mechanism
+
+* Owner: Lei Zhang (@resouer), Jianbo Sun (@wonderflow)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
+
+* Application: A holistic term for talking about all pieces that are rendered from an `ApplicationConfiguration`.
+* ComponentSchematic: A term in OAM spec v1alpha1, which is deprecated and replaced by `Component` in v1alpha2.
+
+## Background
+
+OAM v1alpha1 requires developers to define another ComponentSchematic with
+a different name in order to update it (i.e. ComponentSchematic is immutable). We got feedbacks from
+internal OAM adopters in Alibaba as well as [AWS ECS for OAM](https://github.com/awslabs/amazon-ecs-for-open-application-model#upgrade-and-scale-oam-workloads-with-oam-ecs)
+that it's not a good experience.  Component is now [mutable](https://github.com/oam-dev/spec/pull/356)  in v1alpha2 spec. Here is how we plan to implement it.
+
+Since a component is mutable, we need to address the issue of component versioning.
+
+One of the requirements that we should allow users to specify a fixed version of the component in an ApplicationConfiguration, so that the component will not be upgraded as soon as its definition changes.
+
+On the other hand, the versioning mechanism can also be used by traits. For example, a trait can specify
+a different revision in order to control the upgrade experience. 
+
+
+## Goals
+
+1. Component is mutable. Developers who modify the Component object will directly trigger immediate
+upgrade of any `ApplicationConfiguration` that does not specify `Component` revision. 
+2. Component revision can be specified in ApplicationConfiguration. Operator can specify a fixed revision
+of Component in ApplicationConfiguration. Component mutation won't affect Application directly in this case, 
+and can be upgraded in a more convenient way by working with traits.
+3. The versioning mechanism should be compatible with current existing K8s workloads, eg, `K8s Deployment`, `OpenKruise`.
+  OAM should support two common ways used in community for versioning mechanism. 
+  - Workload itself can handle version. This means we can just update `Deployment` and let the workload do rollout or rollback without traits.
+  - Corporate with traits. OAM runtime should create different `Deployment` instances for different revision and let the traits do rollout or rollback.
+
+## Non-Goals
+
+This proposal will not give a specific rollout or rollback design, that can be done by traits.  
+
+## Proposal
+
+### Using ControllerRevision to record versions of Component
+
+Because a `Component` is mutable, we borrow an independent Revision object called [`ControllerRevision`](https://godoc.org/k8s.io/api/apps/v1#ControllerRevision) from the existing Kubernetes resources.
+to track a `Component` change automatically.
+
+For example, we have a following component yaml:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: Component
+metadata:
+  name: frontend
+spec:
+  workload:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash lscpu"
+```
+
+When users apply this yaml file into the system.
+
+```shell script
+$ kubectl apply -f component.yaml
+```
+
+A `ControllerRevision` object will be created.
+
+```shell script
+$ kubectl get controllerrevisions
+NAMESPACE     NAME                  CONTROLLER               REVISION   AGE
+default       frontend-c8bb659c5    core.oam.dev/component   1          2d15h
+
+$ kubectl get controllerrevisions frontend-c8bb659c5 -o yaml
+apiVersion: apps/v1
+kind: ControllerRevision
+metadata:
+  name: frontend-c8bb659c5 # you could name this anything you wanted, but just appending the semver would be good practice
+revision: 1
+data:
+  workload:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash lscpu"
+```
+
+And it will be recorded in the status of `component.yaml`.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: Component
+metadata:
+  name: frontend
+spec:
+  workload:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash lscpu"
+status:
+  latestRevision: frontend-c8bb659c5
+```
+
+
+If you make a change to `component.yaml`:
+
+```shell script
+           cmd:
+-            - "bash lscpu"
++            - "bash top"
+```
+
+A new `ControllerRevision` will be automatically generated:
+
+```shell script
+$ kubectl get ControllerRevisions
+NAMESPACE     NAME                  CONTROLLER               REVISION   AGE
+default       frontend-c8bb659c5    core.oam.dev/component   1          2d15h
+default       frontend-a75588698    core.oam.dev/component   2          2d14h
+
+$ kubectl get ControllerRevisions frontend-a75588698 -o yaml
+apiVersion: apps/v1
+kind: ControllerRevision
+metadata:
+  name: frontend-a75588698 # you could name this anything you wanted, but just appending the semver would be good practice
+revision: 2
+spec:
+  workload:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash top"
+```
+
+Now with `ControllerRevision` in place, let's talk about how to use it in Trait and ApplicationConfiguration.
+
+### ControllerRevision could be referenced by Trait 
+
+Some traits can control version of component which means they can specify a component revision and won't be affected
+by the change of component.
+
+For these traits, they have to declare `revisionEnabled` field to be `true` in TraitDefinition like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: fancytraits.core.oam.dev
+spec:
+  revisionEnabled: true
+  definitionRef:
+    name: fancytraits.core.oam.dev
+```
+
+Workload instance's name emitted by OAM runtime will be decided by the `revisionEnabled` field.
+
+- If **NO** traits binding with a component has `revisionEnabled` as true, OAM runtime will let the workload itself to handle
+   versioning things. We will only have one workload instance with the same name of componentName, and update workload directly
+   when component changed.
+- If **ANY** trait binding with a component has `revisionEnabled` as true, OAM runtime will always create a new workload with
+   a new name which can be aligned with the name of newly created component revision. OAM runtime will leave the trait to
+   control existing workloads to do rolling update and garbage collection. 
+
+So the essential difference between true or false of `revisionEnabled` is whether we create or update workload.
+
+For example, when we don't use any trait who has revisionEnabled as true, things will work like below:
+
+Firstly, we create an `ApplicationConfiguration` with definition : 
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - componentName: frontend
+```
+
+The real workload instance will be created with the same name as the componentName.
+
+```yaml
+  apiVersion: core.oam.dev/v1alpha2
+  kind: ContainerizedWorkload
+  metadata:
+    name: frontend
+  spec:
+    containers:
+    - name: my-cool-workload
+      image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+      cmd:
+      - "bash lscpu"
+```
+
+But if we use any trait with revisionEnabled.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: FancyTrait
+            spec:
+              traffic:
+                - revisionName: frontend-c8bb659c5
+                # revision: 1  # optional here, which can replace revisionName for UX              
+                  percent: 80%
+                - revision: latest
+                  percent: 20%                 
+```
+
+The real workload instance will be created with the same name as the RevisionName.
+
+```yaml
+  apiVersion: core.oam.dev/v1alpha2
+  kind: ContainerizedWorkload
+  metadata:
+    name: frontend-c8bb659c5
+  spec:
+    containers:
+    - name: my-cool-workload
+      image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+      cmd:
+      - "bash lscpu"
+```
+
+**NOTICE:** In this case, many workload instances of `frontend` will be created by OAM runtime.
+The `FancyTrait` should ensure only the latest workload and the one whose name is `frontend-c8bb659c5` can be running. 
+
+### Always Using the latest revision when using componentName field
+
+When we use a Component with componentName in ApplicationConfiguration, OAM runtime will always use the latest
+component revision to update/create workload and all corresponding traits. 
+
+For example, below is an AppConfig containing one component with two traits.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: rollouts.core.oam.dev
+spec:
+  revisionEnabled: true
+  definitionRef:
+    name: rollouts.core.oam.dev
+---
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: ManuelScaler
+            spec:
+              replica: 3
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: Rollout
+            spec:
+              source:
+                revision: 1
+```
+
+Once the `frontend` component changed, the `ApplicationConfiguration` controller will trigger
+an upgrade for the application.
+
+1. Because the component has a revisionEnabled trait binding with it, so we will aways create new workload instead of update. 
+2. Create a workload according to the new component revision. 
+3. Update all traits pointing their target workloadRef to the new workload.
+4. Rollout trait will do rolling update from `frontend-c8bb659c5`(revision: 1) to the new one, and finally delete the source workload.
+
+
+### Using revisionName field can specify a fixed component version
+
+When a `ControllerRevision` is specified in ApplicationConfiguration like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - revisionName: frontend-c8bb659c5
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: ManuelScaler
+            spec:
+              replica: 3
+```
+
+In this case:
+- The name of workload instance will be the same with revisionName.
+- all traits will point to the specified revision of workload. 
+- Component change won't affect the running application, but `ControllerRevision` will still be created.
+
+Old revisions can also work if any revision aware traits pointing to them.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: traffics.core.oam.dev
+spec:
+  revisionEnabled: true
+  definitionRef:
+    name: traffics.core.oam.dev
+---
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - revisionName: frontend-a75588698
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: Traffic
+            spec:
+              route:
+                - revision: frontend-c8bb659c5 # old revision
+                  percent: 60%
+                - revision: frontend-a75588698 # the new one
+                  percent: 40%
+```
+
+Still, the `Traffic` trait should handle garbage collection things for the old workloads when no longer needed.
+
+## Example: A blue-green workflow
+
+Let's say we have a `Rollout` Trait with definition like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: rollouts.extend.oam.dev
+spec:
+  revisionEnabled: true
+  definitionRef:
+    name: traffics.core.oam.dev
+```
+
+It was used in application with a component.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: extend.oam.dev/v1
+            kind: Rollout
+            spec:
+              replica: 10
+              batch: 3
+              maxUnavailable: 1
+```
+
+Before changing the component, the App will be running in a stable state like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ContainerizedWorkload
+  metadata:
+    name: frontend-c8bb659c5
+spec:
+  containers:
+    - name: my-cool-workload
+      image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+      cmd:
+        - "bash lscpu"
+---
+apiVersion: extend.oam.dev/v1
+kind: Rollout
+spec:
+  batch: 3
+  replica: 10
+  maxUnavailable: 1
+  workloadRef:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    name: frontend-c8bb659c5
+status:
+  currentWorkloadRef:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    name: frontend-c8bb659c5
+```
+
+When `frontend` component changed, new workload will be created with rollout trait updated,
+and the old workload will still be there.
+
+
+```yaml
+# the old one 
+apiVersion: core.oam.dev/v1alpha2
+kind: ContainerizedWorkload
+  metadata:
+    name: frontend-c8bb659c5
+spec:
+  containers:
+    - name: my-cool-workload
+      image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+      cmd:
+        - "bash lscpu"
+---
+# the new one 
+apiVersion: core.oam.dev/v1alpha2
+kind: ContainerizedWorkload
+metadata:
+  name: frontend-a75588698
+spec:
+  containers:
+    - name: my-cool-workload
+      image: example/very-cool-workload:new
+      cmd:
+        - "bash top"
+---
+# rollout spec was pointing to the new workload, but status is still refer to the old.
+apiVersion: extend.oam.dev/v1
+kind: Rollout
+spec:
+  batch: 3
+  replica: 10
+  maxUnavailable: 1
+  workloadRef:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    name: frontend-a75588698
+status:
+  currentWorkloadRef:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    name: frontend-c8bb659c5
+```
+
+Then the rollout trait will take control of the blue-green deploy progress.
+
+Finally the old workload will be removed by rollout trait and the status of rollout trait will align with the spec.
+Getting into another stable state.
+
+
+## Impact to the existing system
+
+1. Add `revisionName` field into AppConfig and this field is mutually exclusive with `componentName`,
+   also `componentName` will be optional if `revisionName` is used.
+2. Add `revisionEnabled` flag into TraitDefinition, and it will affect workload name created by OAM runtime.
+3. A new `ControllerRevision` is created for a `Component` when its parameters changed. A change in `parameterValue`
+   does not affect a `Component` revision. ParameterValue will be configured in ApplicationConfiguration which was supposed
+   to be some kind of operational action, so it will be treated as a special kind of trait. The encouraged usage of
+   parameterValues is for [parameter passing](https://github.com/crossplane/oam-kubernetes-runtime/pull/24).
+   It's an anti-pattern to use parameters to rollout the app (e.g. update image with parameterValue).
+
+
+## Alternative approach
+
+We do discuss a lot for this proposal, most discussion happened in ["OAM component versioning mechanism
+"](https://github.com/oam-dev/spec/issues/336) and ["The Component should mutable"](https://github.com/oam-dev/spec/issues/350).
+Refer to these issues to see more details.


hongchaodeng:
Can you share an example how to do blue-green deployment?

henrywangx:
> Can you share an example how to do blue-green deployment?

Actually, I think it's hard to handle Blue-Green. I'm considering build a workload upon the deployment. Workload revision represents different deployment. That means Blue-Green will build two deployments.

In this way, do canary, blue-green upgrade will be much easier for deployment.

henrywangx:
Meanwhile since for the whole application upgrade, we need to update the component and applicationConfiguration both. For those two config file's changes, it's an async process. So we may generate two revision and trigger 2 times upgrade. How to handle this case?

resouer:
@henrywangx `ParameterValues` is nothing but a special case of Trait, it should not be considered as part of component revision.

You can think of a OAM application is composed by two parts:

1. Data - Components
2. A series of operations to the data - Anything defined in ApplicationConfiguration 

Here is an essential difference between they two: modifying `data` will always trigger a new version of `data`, but modifying `operations` will trigger nothing unless the whole application is re-deployed. For example, imagine Deployment is `data`, and autoscaler is `operation`, modify autoscaler policy should not trigger a new version of Deployment, correct?

`Revision` is used to track history of `data`, because the system need to know what's the current version of `data` it is dealing with, and how to revert to a historical version if needed. In this context, blue-green deployment is very simple in OAM system, just apply a blue-green rollout trait, and modify your Component, a blue-green rollout will happen (rollout trait controller will control the instances of different revision).

For history of operations, the system doesn't care it too much, because even you want to rollback an operation, it's in practical nothing but a re-deploy with new operation policy (this policy is same with a previous operation policy, but the system doesn't care). That being said, you can record the history of operations for audit etc.

henrywangx:
@resouer Yes I understand `ParameterValue` is a special trait like parameterTrait. But modify parameter trait or some other traits like env trait, volumeMount trait may change the workload's spec. That means a new workload revision will generate. Pods with new revision will then generate.

And `autoscaler` is just a special case, it only change the `replica` which won't generate a new ControllerRevision. 

About the Blue-Green upgrade, I guess @hongchaodeng is talking about how a k8s's deployment do the Blue-Green upgrade. From my knowledge, it's really hard to achieve the Blue-Green upgrade for deployment. So I give a solution:

> Actually, I think it's hard to handle Blue-Green. I'm considering to build a workload upon the deployment. Workload revision represents different deployment. That means Blue-Green will build two deployments.
> In this way, do canary, blue-green upgrade will be much easier for deployment.

resouer:
@henrywangx Note that OAM creates its own revision object, it doesn't care revisions generated by Deployment controller. So there's no difference between modifying replicas and patching env list in OAM, all of them will not generate new revision.

> Though we will reuse `ControllerRevision` API in k8s to record the revision, just don't want to reinvent wheel.

I believe Hongchao is talking exactly same idea you are thinking. It's very easy in OAM, you don't need to create a special workload for that:

```
component -> revision-1 --- deployment-1
component change -> revision-2 --- deployment-2
```
so your Blue-green trait controller can always use `revision-1` and `revision-2` to find different deployment object and do rollout.

Also, if another trait, let's say autoscaler changed the replica field of `deployment-1`, it's fine. Because the trait controller only cares that it can find `deployment-1` by given the name of `revision-1`, not the content in the revision object.

btw, the trait controller always handles the lower level object. For example, a autoscaler trait controller will only patch `deployment-1` object, not the `component`. That's why OAM can help to solve the conflicts between human and system which happens a lot in k8s: human only handle OAM objects, system only handle lower level objects.

To ensure this, `oam-kubernetes-runtime` will provide a mechanism (a lib) for any trait controller to be able to locate the Deployment obj with given `ContainerizedWorkload` name. /cc @ryanzhang-oss is working on this. This means even you reference `app.k8s.io/deployment` directly as OAM component, the trait controller will follow the same mechanism to find the Deployment obj.

@wonderflow , maybe code is better to illustrate the whole idea than Github comments 😄

henrywangx:
@resouer thanks for explanation, but I think the [proposal](https://github.com/oam-dev/spec/issues/336#issuecomment-607629006) by @ryanzhang-oss could convince me better. 

There still is some confusion about the proposal, maybe I misunderstand about the idea by @wonderflow . And I will check the code to understand the design.

henrywangx:
After offline discuss with @wonderflow, I understood about the `component->revision->deployment` idea. 
It only treats the component's change as the version change. And the trait's change will be regarded as operations. But in my opinion, user may care more about the final deployment's change(component's change+trait's change) than the component's change.😄

resouer:
> But in my opinion, user may care more about the final deployment's change(component's change+trait's change) than the component's change.

The "final" deployment's state can be always checked by `kubectl get deployment`. 

Though I'm not sure if you get the point of the difference between `data` and `operation`.

Let's say you have a Deployment (replicas=10), and a HPA (min=1;max=100). And at `10:00 am, May 25th, 2020`,  the HPA scaled the Deployment to 20 because its CPU load is high.

Based your prevision suggestion, OAM system will create a "Component Revision (replicas=20)".

But my argument is that "Component Revision (replicas=20)" is actually **useless**. Because `replicas=20` is determined by the CPU load at `10:00 am, May 25th, 2020`. 

So what does it mean if you rollback to `replicas=20`? Can we rollback the CPU load as well 🤔? 

See? The essential difference here is `trait's change` is not determined by trait itself - it's determined by the system.  That's why I said Trait should never be part of `revision` - you can never rollback the `trait's change` unless you can rollback the whole system status.

There's another concept named **snapshot**. For example, you can snapshot all the objects of given `ApplicationConfiguration` every 5 mins. The purpose for snapshot is for **auditing**, and can't rollback (unless you can re-produce the whole system status).

Snapshot is not in the scope of OAM for now.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+upgrade of the application by default. 
+2. Component Version can be specified in ApplicationConfiguration. Operator can specify a fixed version
+of Component in ApplicationConfiguration. Component mutable won't affect Application directly in this case, 
+and can be upgraded in a more convenient way by working with trait.
+
+## Proposal
+
+### Using ControllerRevision to record versions of Component
+
+For component mutable, we introduce an independent Revision object to track Component change automatically.
```
ryanzhang-oss, line 34:
```suggestion
Because a `Component` is mutable, we introduce an independent Revision object to track a `Component` change automatically.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash lscpu"
+```
+
+When user apply this yaml into system, a `ControllerRevision` object will be created.
```
ryanzhang-oss, line 57:
```suggestion
When users apply this yaml file into the system, a `ControllerRevision` object will be created.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+          image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+          cmd:
+          - "bash top"
+```
+
+Now that the basic controller revision are in place, we can use them in Trait and ApplicationConfiguration.
+
+### A real running workload instance will have the same name with revision
+
+I we create an applicationConfiguration like below:
```
ryanzhang-oss, line 144:
```suggestion
If we create an `ApplicationConfiguration` with definition : 
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - componentName: frontend
+```
+
+The real workload instance is created with the same name of the controllerRevision.
```
ryanzhang-oss, line 156:
```suggestion
The real workload instance is created with the same name as the controllerRevision.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+For these traits, they have to declare they are revision aware in TraitDefinition like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: fancytraits.core.oam.dev
+spec:
+  revision:
+    aware: true
```
ryanzhang-oss, line 186: I wonder when do we still  need a  `spec.revision` field when `spec.revision.aware` is false?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+              replica: 3
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: Rollout
+            spec:
+              source:
+                revision: frontend-c8bb659c5 
+```
+
+Once the `frontend` component has been changed, ApplicationConfiguration will trigger
```
ryanzhang-oss, line 255:
```suggestion
Once the `frontend` component is changed, the `ApplicationConfiguration` controller will trigger
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+            kind: Traffic
+            spec:
+              route:
+                - revision: frontend-a1234 # old revision
+                  percent: 60%
+                - revision: frontend-a75588698 # the new one
+                  percent: 40%
+```
+
+We use `[*]` to represent every array in the field.
```
ryanzhang-oss, line 324: is `*` the only quantifier we will support?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+
+
+## Impact to the existing system
+
+1. Add `revisionName` field into AppConfig and this field is mutually exclusive with `componentName`,
+   also `componentName` will be optional if `revisionName` is used.
+2. Add `revision` into TraitDefinition and it will affect workload garbage collection on component upgrade.
+3. Workload instance name will always follow ControllerRevision.
+4. Parameter and ParameterValue need to be treated as a special kind of trait. Values from parameters won't
```
ryanzhang-oss, line 333: I think this proposal is already self-contained. It doesn't conflict with the parameter system which seems to be pretty popular in the OAM community. Namely, a  new `ControllerRevision` is created for a `Component` when its parameters are changed.  A change in `parameterValue` does not affect a `Component` revision. We will address the "uniqueness" of a workload in a different proposal that focuses on the `ApplicationConfiguration` upgrade/verisioning instead of a `Component`.

henrywangx:
@resouer I'm considering the change of `trait` and `parameters` may change the deployment's spec(`replicas` is not in the deployment's spec, yes it depends on the system load). For example, the actor uses `parameters` to replace deployment's image. 

That's why I don't like the paremeters, we have two input paths(component and parameter) to influence the final deployment's spec.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,340 @@
...
+
+For example, below is an AppConfig containing one component with two trait.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: rollouts.core.oam.dev
+spec:
+  revision:
```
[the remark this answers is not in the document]
wonderflow: changed the name, thanks

design/one-pager-component-mutable-and-versioning.md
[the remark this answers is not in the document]
wonderflow: thanks, fixed

wonderflow:
> Can you share an example how to do blue-green deployment?

blue-green example was added

design/one-pager-component-mutable-and-versioning.md
wonderflow: I changed to alwaysControl, and this will have new meanings:

design/one-pager-component-mutable-and-versioning.md
wonderflow: Yes, I think so

design/one-pager-component-mutable-and-versioning.md
wonderflow: Yeah

wonderflow:
After deep consideration for some of your comments and trying to implement this proposal, I  have updated something, please take a second review. Thanks very much.

@henrywangx @resouer @ryanzhang-oss @hongchaodeng

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -183,13 +184,24 @@ metadata:
...
   definitionRef:
     name: fancytraits.core.oam.dev
 ```
 
+#### Revision field rule
+
+Versioning mechanism will be related with `revision` field.
+
+- If no traits binding with a component have revision field, OAM runtime will always update workload directly.
+- If `alwaysControl` field is true, OAM runtime will always create a new workload when component revision changed
```
ryanzhang-oss, line 199: > when component revision changed

I wonder if you can clarify which component revision changed? Is it the revision in the trait or the component itself

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -183,13 +184,24 @@ metadata:
...
+  and leave the trait to control two workloads rolling update.
+- If `alwaysControl` field is false, this means the trait will control only if revision path specified. OAM runtime
```
ryanzhang-oss, line 201: Sounds like we can have a trait definition like below?
```yaml
spec:
    revision:
       alwaysControl: false
```

What does this mean?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -252,16 +264,18 @@ spec:
...
+Once the `frontend` component changed, the `ApplicationConfiguration` controller will trigger
 an upgrade for the application.
 
-1. Create a new workload with the same name of the new component revision.
-2. Check all traits' traitDefinition.
-   - Update all traits who have no revision aware to point to the new workload.
-   - If there are traits who have revision aware and specify to old revision, make sure the workload
+1. Create or update the workload to the new component revision according to revision field of binding traits.
+   - if creating new workload, workload name will align with revision name.
+   - if updating workload, workload name will align with the component name.
```
ryanzhang-oss, line 272: so the workload name will change during updating?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -252,16 +264,18 @@ spec:
...
+   - if a trait who can control revision removed, a new workload will still be created with the same name of component.
```
ryanzhang-oss, line 273: > if a trait who can control revision removed, a new workload will still be created with the same name of component.

I am not entirely sure what does this mean

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,475 @@
...
+  name: example-appconfig
+spec:
+  components:
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: FancyTrait
+            spec:
+              revision: frontend-c8bb659c5
```
ryanzhang-oss, line 220: is the `alwaysControl` field optional?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,475 @@
...
+
+1. Create or update the workload to the new component revision according to revision field of binding traits.
+   - if creating new workload, workload name will align with revision name.
+   - if updating workload, workload name will align with the component name.
+   - if a trait who can control revision removed, a new workload will still be created with the same name of component. 
+2. Update all traits pointing their target workloadRef to the new workload.
+   - If there are traits who have revision field and specify to old revision, make sure the workload
+   corresponding to the old revision will still work. 
+3. Periodically do garbage collection for all workloads which don't have any trait specified.
+4. Traits who have revision field should take care of the garbage collection things for the old revision.
```
ryanzhang-oss, line 278: who is responsible for GC? Traits or AppConfig

design/one-pager-component-mutable-and-versioning.md
henrywangx: > If `alwaysControl` field is true, OAM runtime will always create a new workload when component ~~revision~~ changed

remove revision?
@wonderflow

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,475 @@
...
+    kind: ContainerizedWorkload
+    name: frontend-xas138742
+status:
+  currentWorkloadRef:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    name: frontend-a75588698
+```
+
+Then the rollout trait will take control of the blue-green deploy progress.
```
henrywangx, line 454: If the actor didn't specify the rollout trait who is responsible to control the rollout, by the workload itself?
How about the situation when the trait has been removed when doing rollout?

design/one-pager-component-mutable-and-versioning.md
henrywangx: It should be Component Controller do the GC?. Like the relationship between Deployment Revision and Deployment Controller.

design/one-pager-component-mutable-and-versioning.md
wonderflow: @henrywangx good idea

wonderflow: > I wonder if you can clarify which component revision changed? Is it the revision in the trait or the component itself

component itself @ryanzhang-oss

design/one-pager-component-mutable-and-versioning.md
wonderflow: This means you need to check if the trait specify a revisonPath to know whether we need to create new workload with revision name

design/one-pager-component-mutable-and-versioning.md
wonderflow: yes, according to trait

design/one-pager-component-mutable-and-versioning.md
wonderflow: This case tells what happened if `rollout` trait removed. The workload name will change from revisionName to componentName.

design/one-pager-component-mutable-and-versioning.md
wonderflow: > who is responsible for GC? Traits or AppConfig

When we have revision control trait binding with an App, then Trait is responsible for GC. Otherwise, the AppConfig controller is responsible.

> It should be Component Controller do the GC? Like the relationship between Deployment Revision and Deployment Controller.

Component Controller won't do the GC things.

design/one-pager-component-mutable-and-versioning.md
wonderflow: Yes, it's optional, but you are commenting the wrong place, right? Only traitDefinition has alwaysControl field.

design/one-pager-component-mutable-and-versioning.md
wonderflow: > If the actor didn't specify the rollout trait who is responsible to control the rollout, by the workload itself?

Yes, the workload itself. We really have lots of workloads like Deployment/Statefulset/OpenKruise which have some rollout features.

> How about the situation when the trait has been removed when doing rollout?

First of all, that's an unexpected action, OAM runtime should prevent it happen. If this situation really happen, the AppConfig controller will update the workload and remove the old workload.

resouer:
>  I'm considering the change of trait and parameters may change the deployment's spec

@henrywangx the encouraged usage of parameterValues is for parameter passing, for example: https://github.com/crossplane/oam-kubernetes-runtime/pull/24. It's an anti-pattern to use parameters to rollout the app (update image etc), we will clarify this clearly in the doc after component revision is implemented.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,475 @@
...
+      - "spec.revision"
+  definitionRef:
+    name: fancytraits.core.oam.dev
+```
+
+#### Revision field rule
+
+Versioning mechanism will be related with `revision` field.
+
+- If no traits binding with a component have revision field, OAM runtime will always update workload directly.
```
resouer, line 198: The essential difference is workload CR's name is revisioned or not. The narrative of "update new workload directly" is not accurate.

Also, for both cases, OAM runtime will generate `ControllerRevision` object to track the change of component.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,483 @@
...
+  - Corporate with trait. OAM runtime should create different `Deployment` instances for different revision and let the trait do rollout or rollback.
+
+## Proposal
+
+### Using ControllerRevision to record versions of Component
+
+Because a `Component` is mutable, we introduce an independent Revision object to track a `Component`
+change automatically.
+
+We will use [`ControllerRevision`](https://godoc.org/k8s.io/api/apps/v1#ControllerRevision) in K8s as the Revision object.
```
hasheddan, line 41: I like the idea of using existing k8s primitives here, but I worry that the `ControllerRevision` object is going to be a bit too lightweight for versioning of a `Component`. The `ControllerRevision` object is used by higher-level objects like `DaemonSet` and `StatefulSet` to handle update and rollback, but does not appear to be robust enough for a case like `Component` where an organization may want to "publish" multiple versions of a `Component` that will be consumed by multiple `ApplicationConfigurations`. For instance, if I was authoring a _new_ `ApplicationConfiguration`, specifying a `revisionName` that corresponds to an `int64` `revision` does not feel like a "application owner" level decision.

It seems like what we really want to capture here is versioning of an `ApplicationConfiguration`, for which the `ControllerRevision` feels much more appropriate. For versioning of a `Component`, I think I would actually advocate for making the current notion of a `Component` immutable, such that a developer would create a new `Component` entirely when they want it to be changed. However, this clearly makes it difficult to group `Components` together as they would have to have different names.

A solution to this would be to have a `VersionedComponent` and a `Component` (definitely open to different / better names :sweat_smile: ). The `VersionedComponent` would look very similar to the existing current implementation of a `Component`, but would include a `version` field and a `componentRef` field in the `spec`. The new `Component` would be a vehicle for keeping track of `VersionedComponents`.

Let's explore some tangible examples of what this could look like:
1. Create a `Component` called `frontend`
```yaml
kind: Component
metadata:
  name: frontend
spec:
  # versions will start out as empty
```
2. Create the first version of the `frontend` as a `VersionedComponent`
```yaml
kind: VersionedComponent
metadata:
  name: frontend-v0.0.1 # you could name this anything you wanted, but just appending the semver would be good practice
spec:
  version: v0.0.1
  componentRef:
    name: frontend
  workload:
    kind: ContainerizedWorkload
    spec:
      containers:
      - name: my-cool-workload
        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
        cmd:
        - "bash lscpu"
```
3. A `Component` controller sees this and adds an entry to the `Component` with the `VersionedComponent`:
```yaml
kind: Component
metadata:
  name: frontend
spec:
  versions:
    - v0.0.1: frontend-v0.0.1
```
4a. An `ApplicationConfiguration` is created that references the `Component` and a `version`:
```yaml
metadata:
  name: example-appconfig
spec:
  components:
    - componentName: frontend
      version: v0.0.1 # this would be optional
```
4b. An `ApplicationConfiguration` is created that references a `Component` only:
```yaml
metadata:
  name: example-appconfig
spec:
  components:
    - componentName: frontend # This would result in the latest semantic version being used from the versions map on the Component
```

Now let's look at update strategy. We could do as suggested in the proposal in this PR and if no version is specified then we automatically bump to new versions added to a `Component`. This would be handled by the `ApplicationConfiguration` controller. On each reconcile it would `GET` the `Component` and check to make sure that the latest version is being used. If not, then it will update the rendered `ContainerizedWorkload` in this case. This is where the `ControllerRevision` could come in, if we either automatically update due to a new version being added to the `Component` or manually update due to an explicit `version` change on the `ApplicationConfiguration`, we can create a `ControllerRevision` (or some other new object type if necessary) to handle update and rollback. This makes more sense in my opinion because we are not interested in update and rollback of components themselves, but rather of the _usage_ of components in an `ApplicationConfiguration`.

This would stil allow for usage of traits such as `Rollout` or `TrafficSplit`. but they would only need to specify semantic versions instead of specific revision names.

resouer: Hi Daniel, thanks for detailed review comment. Yes, we agree with you creating a new revision object is neat since k8s' `ControllerRevision` is a generic purpose snapshot object with `data` section, it's not ideal for a user facing revision recorder. I think @wonderflow is more like trying to minimize change to the spec since we don't have a structured way to promote spec change.

So here's my suggestion: 
-  we propose a new revision object in this proposal per your suggestion (need to pick a name: ComponentRevision, VersionedComponent etc);
   - consider: `apiVersion: experimental.oam.dev/v1alpha2`.
- I will try to work out a way to enable implementation driven and versioned spec change promotion mechanism, so we can introduce change safely to spec. 
    - the idea is we will add `experimental` to api version of new object (see above) and new field, see https://github.com/crossplane/oam-kubernetes-runtime/pull/24#issuecomment-634840881. And we will cut release in OAM spec by promoting them to `core.oam.dev/v1alpha3` and remove the `experimental` tag.

Would be glad to know your ideas on this part as well.

Snapshotting ApplicationConfiguration is not in scope for this PR, but I agree we will use ControllerRevision in this case.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,483 @@
...
+      kind: ContainerizedWorkload
+      spec:
+        containers:
+        - name: my-cool-workload
+          image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+          cmd:
+          - "bash top"
+```
+
+Now the basic controller revision are in place, let's talk about them in Trait and ApplicationConfiguration.
```
zzxwill, line 145: The design of `ControllerRevision` is cool. I really needed a versioned component, when I was trying to resolve the [issue](https://github.com/crossplane/provider-alibaba/issues/23) :)
 
```suggestion
Now the basic controller revisions are in place, let's talk about them in Trait and ApplicationConfiguration.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,483 @@
...
+spec:
+  components:
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: FancyTrait
+            spec:
+              traffic:
+                - revision: frontend-c8bb659c5
```
zzxwill, line 219: Agreed with opinions from @hasheddan , I would not prefer to specify a component `reversion` value for an `ApplicationConfiguration`. Especially in a situation when an `ApplicationConfiguration` already exists, I need to modify it or do deletion and recreation operation.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,483 @@
...
+
+OAM v1alpha1 made a requirement that developer need to define another ComponentSchematic with
+different name for an update (i.e. ComponentSchematic is immutable). We've got lots of feedback from
+real practices in Alibaba as well as [AWS ECS for OAM](https://github.com/awslabs/amazon-ecs-for-open-application-model#upgrade-and-scale-oam-workloads-with-oam-ecs)
+that it's not a good experience. Today [Component is mutable in v1alpha2 spec](https://github.com/oam-dev/spec/pull/356),
+this proposal is going to make it into implementation.
+
+Since component is becoming mutable, a very common question is how will the component versioning mechanism work.
+One of the requirement is the versioning mechanism should allow users to specify a fixed version of component
+so that the Application can be upgraded as before. On the other hand, the versioning mechanism should also be used
```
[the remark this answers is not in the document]
wonderflow: In this proposal, component change will trigger an auto upgrade without changing ApplicationConfiguration. But currently, any upgrade happened from the ApplicationConfiguration change. So I means we could specify revision of component to let the upgrade happen as the old way.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,483 @@
...
+
+1. Component is mutable. Developer who modify the Component object will directly trigger further
+upgrade of the application by default. 
+2. Component Version can be specified in ApplicationConfiguration. Operator can specify a fixed version
+of Component in ApplicationConfiguration. Component mutable won't affect Application directly in this case, 
+and can be upgraded in a more convenient way by working with trait.
+3. The versioning mechanism should compatible with current existing K8s workloads, eg, `K8s Deployment`, `OpenKruise`.
+  OAM should support two common ways used in community for versioning mechanism. 
+  - Workload itself can handle version. This means we can use `Deployment` only to do rollout or rollback without trait.
+  - Corporate with trait. OAM runtime should create different `Deployment` instances for different revision and let the trait do rollout or rollback.
```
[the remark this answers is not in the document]
wonderflow: Your understand is correct. 

> it may exist simultaneously with the previous one

But only one of the case will happen at the same time. It's depending on the trait.

design/one-pager-component-mutable-and-versioning.md
wonderflow: Yeah, I'm agree with using `ComponentRevision` instead of `ControllerRevision`. But for this:

```
metadata:
  name: example-appconfig
spec:
  components:
    - componentName: frontend
      version: v0.0.1 # this would be optional
```

I think we can still use `revisionName` to do so, since we have [discussed a lot](https://github.com/oam-dev/spec/issues/350)  before and have already [merged in spec](https://github.com/oam-dev/spec/pull/356). They are almost the same.

design/one-pager-component-mutable-and-versioning.md
wonderflow: see https://github.com/crossplane/oam-kubernetes-runtime/pull/29#discussion_r432265873

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,483 @@
...
+so that the Application can be upgraded as before. On the other hand, the versioning mechanism should also be used
+by traits. For example, a trait can specify different revision and control deployment process. 
+
+
+## Goals
+
+1. Component is mutable. Developer who modify the Component object will directly trigger further
+upgrade of the application by default. 
+2. Component Version can be specified in ApplicationConfiguration. Operator can specify a fixed version
+of Component in ApplicationConfiguration. Component mutable won't affect Application directly in this case,
```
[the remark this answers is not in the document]
wonderflow: yes, added

wonderflow:
@hasheddan @zzxwill  Thanks so much for your review, I think we're all agree with `ComponentRevision` and the revisionName was already merged in OAM spec so I hope we can follow that. All of the rest comments has been updated. 

I'm going to implement this proposal if there are no other different ideas. \cc @resouer @hongchaodeng @ryanzhang-oss @henrywangx

vturecek:
@hasheddan made an interesting suggestion to introduce versioned components, where each version is immutable. That makes sense to me, as immutability to me simply means that a component config is unique to a name-version pair. An Application Configuration should _always_ be specific about the version of a component that's being run. I think it would be a huge mistake to allow the equivalent of the "latest" tag for container images with Component instances.

resouer:
@vturecek This proposal is stick to have `revisionName` in `ApplicationConfiguration` which is based on the conclusion during previous discussion: https://github.com/oam-dev/spec/issues/350#issuecomment-617556483. And as you suggested, the `revisionName` is encouraged approach.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+OAM v1alpha1 made a requirement that developer need to define another ComponentSchematic with
+different name for an update (i.e. ComponentSchematic is immutable). We've got lots of feedback from
+real practices in Alibaba as well as [AWS ECS for OAM](https://github.com/awslabs/amazon-ecs-for-open-application-model#upgrade-and-scale-oam-workloads-with-oam-ecs)
+that it's not a good experience. Today [Component is mutable in v1alpha2 spec](https://github.com/oam-dev/spec/pull/356),
+this proposal is going to make it into implementation.
+
+Since component is becoming mutable, a very common question is how will the component versioning mechanism work.
+
+One of the requirement is the versioning mechanism should allow users to specify a fixed version of component
+in ApplicationConfiguration, so that the component can be upgraded as before.
```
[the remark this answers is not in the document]
resouer: > Should this requirement be re-examined instead? Does using Helm solve this problem adequately? Or do we think that version/revision tracking is fundamentally necessary for some of the upgrade features we want to add?

So the intention of having revision in OAM is to let Trait have a way to reference revisions when needed, for example, A/B testing trait and traffic splitting trait. AppConfig will not handle the upgrade story, it's only responsible for "emitting" correct revision of workload CR to k8s.

> It seems complex understand that ApplicationConfiguration both supports snapshotting (explicit versions everywhere) and automatic upgrade (no versions specified).

Yes, though it should be implementable. This is a compromise that both scenarios are valid and widely used in the community.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+```shell script
+$ kubectl get componentrevisions
+NAMESPACE     NAME                  CONTROLLER               REVISION   AGE
+default       frontend-v0.0.1    core.oam.dev/component   1          2d15h
+
+$ kubectl get componentrevisions frontend-v0.0.1 -o yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ComponentRevision
+metadata:
+  name: frontend-v0.0.1 # you could name this anything you wanted, but just appending the semver would be good practice
```
artursouza, line 85: Name should not include version. These are two different things. Runtime should handle version separately.

artursouza, line 85: version should be here and not in spec.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+spec:
+  components:
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: FancyTrait
+            spec:
+              traffic:
+                - revision: frontend-v0.0.2
```
artursouza, line 213: this is not revision, it is name + revision. Shouldn't this be separate fields?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+            kind: Rollout
+            spec:
+              source:
+                revision: frontend-v0.0.1 
+```
+
+Once the `frontend` component changed, the `ApplicationConfiguration` controller will trigger
+an upgrade for the application.
+
+1. Because the component has a revisionAware trait binding with it, so we will aways create new workload instead of update.
```
artursouza, line 278: Can the attribute mean update vs create instead of "revisionAware"?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+When a `ComponentRevision` is specified in ApplicationConfiguration like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - revisionName: frontend-v0.0.1
```
artursouza, line 295: Can we use two attributes instead: name and version?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+
+Then the rollout trait will take control of the blue-green deploy progress.
+
+Finally the old workload will be removed by rollout trait and the status of rollout trait will align with the spec.
+Getting into another stable state.
+
+
+## Impact to the existing system
+
+1. Add `revisionName` field into AppConfig and this field is mutually exclusive with `componentName`,
```
artursouza, line 463: Can we just get away by adding "version" to Component?

design/one-pager-component-mutable-and-versioning.md
wonderflow: Do you mean, name should be something like  `frontend-adhsyradsx` hash string?

design/one-pager-component-mutable-and-versioning.md
wonderflow: Is `frontend-c8bb659c5` match?

design/one-pager-component-mutable-and-versioning.md
wonderflow: Yes, it means update vs create. Do you have better word to represent it?

design/one-pager-component-mutable-and-versioning.md
wonderflow: These two comments above point to the same thing, and it's already discussed here https://github.com/oam-dev/spec/issues/350#issuecomment-617556483.  I'm following the conclusion.

design/one-pager-component-mutable-and-versioning.md
wonderflow: same  with https://github.com/crossplane/oam-kubernetes-runtime/pull/29#discussion_r433637657

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+* Owner: Lei Zhang (@resouer), Jianbo Sun (@wonderflow)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
+
+* Application: A holistic term for talking about all pieces that are rendered from an `ApplicationConfiguration`.
+* ComponentSchematic: A term in OAM spec v1alpha1, which is deprecated and replaced by `Component` in v1alpha2.
+* ComponentRevision: A new concept introduced in this proposal which will work as revision of component.  
+
```
zzxwill, line 12: Will you explain a little bit about `ControllerRevision` in this `Terminology` section?

design/one-pager-component-mutable-and-versioning.md
resouer: +1, revision name should be unique and auto generated by system: `frontend-578mv`

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+default       frontend-v0.0.1    core.oam.dev/component   1          2d15h
+
+$ kubectl get componentrevisions frontend-v0.0.1 -o yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ComponentRevision
+metadata:
+  name: frontend-v0.0.1 # you could name this anything you wanted, but just appending the semver would be good practice
+spec:
+  version: v0.0.1
+  componentRef:
```
resouer, line 88: The owner of this `ComponentRevision` is the frontend component, so no need for this field.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+NAMESPACE     NAME                  CONTROLLER               REVISION   AGE
+default       frontend-v0.0.1    core.oam.dev/component   1          2d15h
+
+$ kubectl get componentrevisions frontend-v0.0.1 -o yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ComponentRevision
+metadata:
+  name: frontend-v0.0.1 # you could name this anything you wanted, but just appending the semver would be good practice
+spec:
+  version: v0.0.1
```
resouer, line 87: There's no need to explicitly claim a `version` IMO. Revision object should be generated by the system, so it doesn't make lots sense to force the system generate version number incrementally like this.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,477 @@
...
+spec:
+  workload:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash lscpu"
```
resouer, line 67: Here's a missing part: where will be the latest revision of this Component be recorded? Annotation? Or Status?

I'd prefer status since it's structured, for example:

```yaml
status:
  latestCreatedRevisionName: frontend-578mv
```

design/one-pager-component-mutable-and-versioning.md
wonderflow: I'm fine but what's the difference between controllerRevision with  ComponentRevision if we remove this version field ?

resouer: The main difference is `ControllerRevision` is not human readable as it uses a generic `data` field to carry the revisioned object.

So if the argument is user experience, `ComponentRevision` proposed in https://github.com/crossplane/oam-kubernetes-runtime/pull/29#discussion_r431915520 is better - still, not a very big difference indeed. 

While what I am not sure on Daniel is comment is:

> but does not appear to be robust enough for a case like Component where an organization may want to "publish" multiple versions of a Component that will be consumed by multiple ApplicationConfigurations

I don't think it's the intention in this proposal. Revision is OAM is mostly for AppConfig/Trait controller to find workload CR by given revision number.

design/one-pager-component-mutable-and-versioning.md
wonderflow: yes, I will add in status

wonderflow:
@hasheddan  Finally I decide to use controllerRevision instead of creating a new concept `componentRevision`, because they have almost the same experience for user.

wonderflow:
@resouer @artursouza  I think I have fixed all comments and already have an implementation(https://github.com/crossplane/oam-kubernetes-runtime/pull/35). Can we approve this proposal and let's move forward quickly?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+* Application: A holistic term for talking about all pieces that are rendered from an `ApplicationConfiguration`.
+* ComponentSchematic: A term in OAM spec v1alpha1, which is deprecated and replaced by `Component` in v1alpha2.
+
+## Background
+
+OAM v1alpha1 made a requirement that developer need to define another ComponentSchematic with
+different name for an update (i.e. ComponentSchematic is immutable). We've got lots of feedback from
+real practices in Alibaba as well as [AWS ECS for OAM](https://github.com/awslabs/amazon-ecs-for-open-application-model#upgrade-and-scale-oam-workloads-with-oam-ecs)
+that it's not a good experience. Today [Component is mutable in v1alpha2 spec](https://github.com/oam-dev/spec/pull/356),
+this proposal is going to make it into implementation.
```
ryanzhang-oss, line 18:
```suggestion
OAM v1alpha1 requires developers to define another ComponentSchematic with
a different name in order to update it (i.e. ComponentSchematic is immutable). We got feedbacks from
internal OAM adopters in Alibaba as well as [AWS ECS for OAM](https://github.com/awslabs/amazon-ecs-for-open-application-model#upgrade-and-scale-oam-workloads-with-oam-ecs)
that it's not a good experience.  Component is now [mutable](https://github.com/oam-dev/spec/pull/356)  in v1alpha2 spec. Here is how we plan to implement it.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+
+Since component is becoming mutable, a very common question is how will the component versioning mechanism work.
```
ryanzhang-oss, line 20:
```suggestion
Since a component is mutable, we need to address the issue of component versioning.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+
+One of the requirement is the versioning mechanism should allow users to specify a fixed version of component
+in ApplicationConfiguration, so that the component can be upgraded as before.
```
ryanzhang-oss, line 23:
```suggestion
One of the requirements that we should allow users to specify a fixed version of the component in an ApplicationConfiguration, so that the component will not be upgraded as soon as its definition changes.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+
+On the other hand, the versioning mechanism should also be used by traits. For example, a trait can specify
+different revision and control deployment process.
```
ryanzhang-oss, line 26:
```suggestion
On the other hand, the versioning mechanism can also be used by traits. For example, a trait can specify
a different revision in order to control the upgrade experience. 
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+
+## Non-Goals
+
+This proposal will not give a specific rollout or rollback design, that can be done by traits.  
+
+## Proposal
+
+### Using ControllerRevision to record versions of Component
+
+Because a `Component` is mutable, we introduce an independent Revision object called [`ControllerRevision`](https://godoc.org/k8s.io/api/apps/v1#ControllerRevision)
```
ryanzhang-oss, line 49:
```suggestion
Because a `Component` is mutable, we borrow an independent Revision object called [`ControllerRevision`](https://godoc.org/k8s.io/api/apps/v1#ControllerRevision) from the existing Kubernetes resources.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+to track a `Component` change automatically. It is a native K8s resource.
```
ryanzhang-oss, line 50:
```suggestion
to track a `Component` change automatically.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash top"
+```
+
+Now the basic `ControllerRevision` is in place, let's talk about it in Trait and ApplicationConfiguration.
```
ryanzhang-oss, line 158:
```suggestion
Now with `ControllerRevision` in place, let's talk about how to use it in Trait and ApplicationConfiguration.
```

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+  createRevision: true
+  definitionRef:
+    name: fancytraits.core.oam.dev
+```
+
+Workload instance's name emitted by OAM runtime will be decided by the `createRevision` field.
+
+- If **NO** traits binding with a component has `createRevision` as true, OAM runtime will let the workload itself to handle
+   versioning things. We will only have one workload instance with the same name of componentName, and update workload directly
+   when component changed.
```
ryanzhang-oss, line 182: what about components that specified with `revisionName `?

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,499 @@
...
+                revision: 1
+```
+
+Once the `frontend` component changed, the `ApplicationConfiguration` controller will trigger
+an upgrade for the application.
+
+1. Because the component has a createRevision trait binding with it, so we will aways create new workload instead of update. 
+2. Create a workload according to the new component revision. 
+3. Update all traits pointing their target workloadRef to the new workload.
+4. Rollout trait will do rolling update from `frontend-c8bb659c5`(revision: 1) to the new one, and finally delete the source workload.
```
ryanzhang-oss, line 303: what happens after the first rollout and the revision 1 is deleted? What does the rollout trait do when the component is changed again?

design/one-pager-component-mutable-and-versioning.md
wonderflow: removed

design/one-pager-component-mutable-and-versioning.md
wonderflow: It will be fixed in that revision, I explained it in latter section

design/one-pager-component-mutable-and-versioning.md
wonderflow: 1. spec.workloadRef will be the same with status, the rollout trait will be in stable state.
2. spec.workloadRef of the rollout trait will change, it will differ from status and the rollout trait will work again.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,497 @@
...
+
+For these traits, they have to declare `createRevision` field to be `true` in TraitDefinition like below:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: fancytraits.core.oam.dev
+spec:
+  createRevision: true
```
hongchaodeng, line 171: `createRevision` is a procedural name. Maybe rename to `revisionEnabled`

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -0,0 +1,497 @@
...
+    kind: ContainerizedWorkload
+    spec:
+      containers:
+      - name: my-cool-workload
+        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
+        cmd:
+        - "bash top"
+```
+
+Now with `ControllerRevision` in place, let's talk about how to use it in Trait and ApplicationConfiguration.
```
hongchaodeng, line 156: How does Workload CR gets created in this case?

design/one-pager-component-mutable-and-versioning.md
wonderflow: i'm fine

hongchaodeng:
lgtm

PR #33, opened 2020-05-29
add OAM policy design

ryanzhang-oss:
Add an initial OAM Policy design

### DIFF
--- a/design/one-pager-oam-policy.md
+++ b/design/one-pager-oam-policy.md
@@ -1,0 +1,243 @@
+# Introduce workload policy in OAM
+
+* Owner: Ryan Zhang (@ryanzhang-oss)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
+
+* **CRD (Custom Resource Definition)** : A standard Kubernetes Custom Resource Definition
+* **CR (Custom Resource)** : An instance of a Kubernetes type that was defined using a CRD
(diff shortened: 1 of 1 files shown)


design/one-pager-oam-policy.md
```diff
@@ -0,0 +1,243 @@
...
+           image: wordpress:latest
+```
+The problem is two folds
+1. Developers may need to express opinions on how to run the workload. We currently do not have a
+ way to express that.
+   - For example,  let's say that the `example-db` component developer wants to deploy the
+    database component in a master only mode. Therefore, an application operator cannot apply a
+    `ManualScalerTrait` to the `example-db` component. In the same time, let's say that the
+    `example-web` component developer wants to expose the component to the internet. This
+     means that an application operator needs to know that it should apply an `IngressTrait' to
```
wonderflow, line 92:
```suggestion
     means that an application operator needs to know that it should apply an `IngressTrait` to
```

design/one-pager-oam-policy.md
```diff
@@ -0,0 +1,243 @@
...
+     kind: ContainerizedWorkload
+     metadata:
+       name: mydb-example
+     spec:
+       containers:
+         - name: mysql
+           image: mysql:latest
+  policies
+    - type: maxInstancePolicy
+      value: 1
```
wonderflow, line 156:
```
- type: replicaPolicy
   value: singleton
```

hongchaodeng:
I think this proposal is missing the problem background and use cases. Would like to see them added first.

PR #37, opened 2020-06-08
Allow trait metadata name to be nil

### LINKED ISSUE #36: Trait name should be automatically generated if empty
image

I know specify a name can allow traits exist more than once for a component. And I also see some cases (e.g. Volume trait. Service trait) that may need more than once. But trait is designed as attribute of component, an attribute should exist only once. 

Before we have a way to match both of the cases,  I suggest we support both of them.  
- If trait name wasn't specified, the name of trait can be automatically generated.
- If trait name already exist, we don't override.

zzxwill:
implement TODO(negz): This name can be omitted and generated
automatically if each trait kind may apply only once to a
component/workload.

Fix #36

### DIFF
--- a/pkg/controller/oam/applicationconfiguration/render.go
+++ b/pkg/controller/oam/applicationconfiguration/render.go
@@ -102,6 +102,11 @@ func (r *components) Render(ctx context.Context, ac *v1alpha2.ApplicationConfigu
 				return nil, errors.Wrapf(err, errFmtRenderTrait, acc.ComponentName)
 			}
 
+			// Set metadata name for `Trait` if the metadata name is NOT set.
+			if t.GetName() == "" {
+				t.SetName(acc.ComponentName)
+			}
+
 			t.SetOwnerReferences([]metav1.OwnerReference{*ref})
 			t.SetNamespace(ac.GetNamespace())
 
--- a/pkg/controller/oam/applicationconfiguration/render_test.go
+++ b/pkg/controller/oam/applicationconfiguration/render_test.go
@@ -442,3 +442,102 @@ func TestResolveParams(t *testing.T) {
 		})
 	}
 }
+
+func TestRenderTraitWithoutMetadataName(t *testing.T) {
+	namespace := "ns"
+	acName := "coolappconfig"
+	acUID := types.UID("definitely-a-uuid")
+	componentName := "coolcomponent"
+	workloadName := "coolworkload"
+
+	ac := &v1alpha2.ApplicationConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      acName,
+			UID:       acUID,
+		},
+		Spec: v1alpha2.ApplicationConfigurationSpec{
+			Components: []v1alpha2.ApplicationConfigurationComponent{
+				{
+					ComponentName: componentName,
+					Traits:        []v1alpha2.ComponentTrait{{}},
+				},
+			},
+		},
+	}
+
+	ref := metav1.NewControllerRef(ac, v1alpha2.ApplicationConfigurationGroupVersionKind)
+
+	type fields struct {
+		client   client.Reader
+		params   ParameterResolver
+		workload ResourceRenderer
+		trait    ResourceRenderer
+	}
+	type args struct {
+		ctx context.Context
+		ac  *v1alpha2.ApplicationConfiguration
+	}
+	type want struct {
+		w   []Workload
+		err error
+	}
+	cases := map[string]struct {
+		reason string
+		fields fields
+		args   args
+		want   want
+	}{
+		"Success": {
+			reason: "One workload and one trait should successfully be rendered",
+			fields: fields{
+				client: &test.MockClient{MockGet: test.NewMockGetFn(nil)},
+				params: ParameterResolveFn(func(_ []v1alpha2.ComponentParameter, _ []v1alpha2.ComponentParameterValue) ([]Parameter, error) {
+					return nil, nil
+				}),
+				workload: ResourceRenderFn(func(_ []byte, _ ...Parameter) (*unstructured.Unstructured, error) {
+					w := &unstructured.Unstructured{}
+					w.SetName(workloadName)
+					return w, nil
+				}),
+				trait: ResourceRenderFn(func(_ []byte, _ ...Parameter) (*unstructured.Unstructured, error) {
+					t := &unstructured.Unstructured{}
+					return t, nil
+				}),
+			},
+			args: args{ac: ac},
+			want: want{
+				w: []Workload{
+					{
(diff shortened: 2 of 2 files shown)


zzxwill:
@wonderflow Please help look into this implementation. 
By this pull request, 
- the `ApplicationConfiguration` in [examples/containerized-workload/sample_application_config.yaml](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/examples/containerized-workload/sample_application_config.yaml) could be deployed.
- component/pod seems normal.
```
➜  /Users/zhouzhengxi/Programming/golang/src/github.com/zzxwill/oam-kubernetes-runtime git:(trait-metadata-name) k get component
NAME                WORKLOAD-KIND
example-component   ContainerizedWorkload
➜  /Users/zhouzhengxi/Programming/golang/src/github.com/zzxwill/oam-kubernetes-runtime git:(trait-metadata-name) k get pod
NAME                                          READY   STATUS    RESTARTS   AGE
example-appconfig-workload-79ffc98486-hr44k   1/1     Running   0          17m
```
- It can allow metadata name of trait to be nil if the name isn't set, and letting it be if set. And the name can show correctly.
image

But yet, I don't know how to get rid of the error message in statue conditions.
image

zzxwill:
There is an ugly workaround to getting rid of the error message by `kubectl edit applicationconfiguration example-appconfig` to manually set the metadata name.

wonderflow:
I didn't reproduce this issue, please add a unit/e2e test for this.

zzxwill:
> I didn't reproduce this issue, please add a unit/e2e test for this.

Done.
Here is the result of `make test`.
```
make test
12:28:21 [ .. ] verify dependencies have expected content
all modules verified
12:28:27 [ OK ] go modules dependencies verified
12:28:27 [ .. ] go test unit-tests
?   	github.com/crossplane/oam-kubernetes-runtime/pkg/controller	[no test files]
?   	github.com/crossplane/oam-kubernetes-runtime/pkg/controller/oam	[no test files]
ok  	github.com/crossplane/oam-kubernetes-runtime/pkg/controller/oam/applicationconfiguration	1.573s	coverage: 95.7% of statements
?   	github.com/crossplane/oam-kubernetes-runtime/pkg/oam	[no test files]
?   	github.com/crossplane/oam-kubernetes-runtime/pkg/oam/fake	[no test files]
?   	github.com/crossplane/oam-kubernetes-runtime/pkg/reconciler	[no test files]
ok  	github.com/crossplane/oam-kubernetes-runtime/pkg/reconciler/trait	0.519s	coverage: 82.9% of statements
ok  	github.com/crossplane/oam-kubernetes-runtime/pkg/reconciler/workload	0.976s	coverage: 84.5% of statements
?   	github.com/crossplane/oam-kubernetes-runtime/pkg/trait	[no test files]
ok  	github.com/crossplane/oam-kubernetes-runtime/pkg/trait/manualscaler	2.037s	coverage: 100.0% of statements
ok  	github.com/crossplane/oam-kubernetes-runtime/pkg/workload	0.538s	coverage: 85.7% of statements
ok  	github.com/crossplane/oam-kubernetes-runtime/pkg/workload/containerized	0.524s	coverage: 42.1% of statements
?   	github.com/crossplane/oam-kubernetes-runtime/apis	[no test files]
?   	github.com/crossplane/oam-kubernetes-runtime/apis/core	[no test files]
?   	github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2	[no test files]
12:28:47 [ OK ] go test unit-tests
```

As `applicatonconfiguration_test.go` won't call render and apply module, I didn't find a way to write a e2e test and just append a unit-test here.

zzxwill:
Besides, thank you for the verification. It didn't reproduce as I tried it another Kubernetes cluster from scratch.

zzxwill:
@wonderflow I get to know e2e test could be write to [test/integration/appconfig_test.go](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/test/integration/appconfig_test.go). Thank you.

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -102,6 +102,11 @@ func (r *components) Render(ctx context.Context, ac *v1alpha2.ApplicationConfigu
 				return nil, errors.Wrapf(err, errFmtRenderTrait, acc.ComponentName)
 			}
 
+			// Set metadata name for `Trait` if the metadata name is NOT set.
+			if t.GetName() == "" {
+				t.SetName(acc.ComponentName)
+			}
```
negz, line 108: What happens if two AppConfigs use the same Component? Won't we end up with multiple traits trying to use the same name?

pkg/controller/oam/applicationconfiguration/render_test.go
```diff
@@ -442,3 +442,102 @@ func TestResolveParams(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			r := &components{tc.fields.client, tc.fields.params, tc.fields.workload, tc.fields.trait}
+			got, _ := r.Render(tc.args.ctx, tc.args.ac)
+			if len(got) == 0 || len(got[0].Traits) == 0 || got[0].Traits[0].GetName() != componentName {
```
negz, line 537: We typically use https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc for comparisons - it's pretty handy and common in Go tests.

pkg/controller/oam/applicationconfiguration/render.go
wonderflow: As discussed before, Component is an instance, one component should be in one appconfig 
 only.

pkg/controller/oam/applicationconfiguration/render_test.go
wonderflow: agree, we will align to that next time, thanks!

pkg/controller/oam/applicationconfiguration/render.go
resouer: > As discussed before, Component is an instance, one component should be in one appconfig
only.

Not sure if we have admission to enforce this practice?

wonderflow: What about use  componentName+<hash-suffix> to fix this case? I don't like appconfigName+componentName because it could be too long. @negz @resouer

wonderflow: ref to admission controller issue https://github.com/crossplane/crossplane/issues/1509

resouer: The current name is fine to me, I'm just suggesting we may need to include admission to enforce this.

PR #46, opened 2020-06-15
add workloadRefPath to traitDefinition

ryanzhang-oss:
Finalize the trait/workoad interaction design. We need to rely on the trait owner to signal if we can add the workloadRef field to its CR. It's very hard for appConfig controller to validate otherwise because

1. CRDs (such as all the existing OAM CRDs in crossplane) may not come with a strict validation schema
2. The trait can come with a strict validation webhook (i.e Knative route) that we have no way to know.

### DIFF
--- a/design/one-pager-trait-workload-interaction-mechanism.md
+++ b/design/one-pager-trait-workload-interaction-mechanism.md
@@ -116,31 +116,30 @@ apply to in the future.
           awsSecret: <aws-secret>
       ```
 
-
 ## Proposal
 The overall idea is for the applicationConfiguration controller to fill critical information
 in the workload and trait CR it emits. In addition, we will provide a helper library so that
 trait controller developers can locate the resources they need with a simple function call. 
 Here is the list of changes that we propose.
-1. Add an optional field called `nativeTrait` to the `traitDefinition` schema. This is for the trait
- owner to declare that a trait is native to the OAM trait system instead of importing from an
- existing operator. This field indicates that the trait owner can modify the trait's operator. It
- implies that the trait relies on the OAM trait/workload interaction mechanism. In our example
- , the trait definition would look like below.  
+1. Add an optional field called `workloadRefPath` to the `traitDefinition` schema. This is for the 
+ trait owner to declare that the trait relies on the OAM trait/workload interaction
+ mechanism. The value of the field is the path to the field that takes a `workloadRef` object. In
+ our example, the trait definition would look like below since our `manualscalertraits` takes
+ the `workloadRef` field at `spec.workloadRef`.
      ```yaml
        apiVersion: core.oam.dev/v1alpha2
        kind: TraitDefinition
        metadata:
          name: manualscalertraits.core.oam.dev
        spec:
-         nativeTrait: true
+         workloadRefPath: spec.workloadRef
          definitionRef:
            name: manualscalertraits.core.oam.dev
      ```
 2. ApplicationConfig controller no longer assumes that all `trait` CRDs contain a "spec
 .workloadRef" field conforms to the OAM definition. It only fills the workload GVK to a `trait
-` CR with "spec.workloadRef" field defined as below if the corresponding `traitDefiniton` is
- marked as "oamAware".   
+` CR with "spec.workloadRef" field defined as below if the corresponding `traitDefiniton` has a
+ "workloadRefPath" field.   
      ```yaml
        workloadRef:
          properties:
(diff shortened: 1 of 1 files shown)


design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -122,9 +122,23 @@ The overall idea is for the applicationConfiguration controller to fill critical
 in the workload and trait CR it emits. In addition, we will provide a helper library so that
 trait controller developers can locate the resources they need with a simple function call. 
 Here is the list of changes that we propose.
-1. ApplicationConfig controller no longer assumes that all `trait` CRDs contain a "spec
+1. Add an optional field called `oamAware` to the `traitDefinition` schema. The trait owner needs
```
wonderflow, line 125: The name `oamAware` is not intuitive, what about "workloadDefNeeded"

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -122,9 +122,23 @@ The overall idea is for the applicationConfiguration controller to fill critical
...
+         name: manualscalertraits.core.oam.dev
+       spec:
+         oamAware: true
+         definitionRef:
+           name: manualscalertraits.core.oam.dev
+     ```
+2. ApplicationConfig controller no longer assumes that all `trait` CRDs contain a "spec
 .workloadRef" field conforms to the OAM definition. It only fills the workload GVK to a `trait
-` CR if its CRD has a "spec.workloadRef" field defined as below.  
+` CR with "spec.workloadRef" field defined as below if the corresponding `traitDefiniton` is
```
zzxwill, line 140: +1

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: I am open to naming. However, the reason for this name is we may reuse this field for future purposes, not limited to workload. So this is intended to be a general-purpose field.

resouer: Yeah, this is a bad name.

Actually, I would suggest change `workloadRef` to `targetRef`, and change this flag to `TargetAware`.

Note that `targetRef` also aligns with what Kubernetes is using today.

hongchaodeng:
What would OAM runtime do if a trait signals "no"?

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: @resouer can you point me to where `targetRef` is used in K8s?

ryanzhang-oss:
> What would OAM runtime do if a trait signals "no"?

The default s "no" so there is no need for a trait to signal anything.  Also, the app Config controller will not patch the `workloadRef` field to a trait by default.

design/one-pager-trait-workload-interaction-mechanism.md
resouer: Check this one: https://cloud.google.com/kubernetes-engine/docs/concepts/verticalpodautoscaler k8s has a built-in type for it.

wonderflow:
@zzxwill  checked that dry-run=Server will work for checking if trait CR contains `workloadRef` field， which we don't need `oamAware` then.

wonderflow:
LGTM

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -179,10 +196,10 @@ deployment and service child resources.
 ## Impact to the existing system
 Here are the impacts of this mechanism to the existing OAM components
 - ApplicationConfiguration: This mechanism requires minimum changes in the
- applicationConfiguration controller.
+ applicationConfiguration controller except that it now needs to check if a trait is "oamAware".
```
hongchaodeng, line 199: > trait is "oamAware"
Please change this too

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -122,9 +122,25 @@ The overall idea is for the applicationConfiguration controller to fill critical
 in the workload and trait CR it emits. In addition, we will provide a helper library so that
 trait controller developers can locate the resources they need with a simple function call. 
 Here is the list of changes that we propose.
-1. ApplicationConfig controller no longer assumes that all `trait` CRDs contain a "spec
+1. Add an optional field called `nativeTrait` to the `traitDefinition` schema. This is for the trait
```
hongchaodeng, line 125: nativeTrait is an ambiguous name.
It is very hard for a user to correlate "native" to the implementation of OAM trait/workload interaction mechanism.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -122,9 +122,25 @@ The overall idea is for the applicationConfiguration controller to fill critical
...
+       spec:
+         nativeTrait: true
+         definitionRef:
+           name: manualscalertraits.core.oam.dev
+     ```
+2. ApplicationConfig controller no longer assumes that all `trait` CRDs contain a "spec
 .workloadRef" field conforms to the OAM definition. It only fills the workload GVK to a `trait
-` CR if its CRD has a "spec.workloadRef" field defined as below.  
+` CR with "spec.workloadRef" field defined as below if the corresponding `traitDefiniton` is
+ marked as "oamAware".
```
hongchaodeng, line 143: This is "oamAware"?

design/one-pager-trait-workload-interaction-mechanism.md
hongchaodeng, line 199: > trait is "oamAware"

Ditto. Needs to update.

design/one-pager-trait-workload-interaction-mechanism.md
resouer: @ryanzhang-oss I will eat my word on the field name. `targetRef` is overlapping. Let's keep `workloadRef`.

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: any suggestions?

PR #49, opened 2020-06-24
auto binding design

wonderflow:
This is a supplemental proposal for [Resource Dependency in OAM](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/design/resource-dependency.md)

### DIFF
--- a/design/auto-binding.md
+++ b/design/auto-binding.md
@@ -1,0 +1,209 @@
+# Auto Binding in OAM
+
+* Owner: Jianbo Sun (@wonderflow)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+[Resource Dependency in OAM](resource-dependency.md) can handle one component depends on another.
+It also contains a basic solution to pass through data from one component to another using
+`dataOutputs` and `dataInputs`. The `toFieldPaths` field of `dataInputs` can specify a fieldPath
+of workload to bind with. But if we want to bind with an object like K8s `Secret` or `ConfigMap`,
+we should bind with the key of these objects to give better user experience.
+
+## Use Case
+
+### Web app and database
+
+Let's use web app and database use case as an example.
+
+We have a db component like below.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: Component
+metadata:
+  name: db
+spec:
+  workload:
+    apiVersion: database.alibaba.crossplane.io/v1alpha1
+    kind: RDSInstance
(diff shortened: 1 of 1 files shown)


SevenYoung:
LGTM

design/auto-binding.md
```diff
@@ -0,0 +1,199 @@
...
+In this case, Developer of web component don't need to specify name of secret, but he has to specify the key of secret.
+
+But if the developer don't know the secret name, how can he know the key? So I propose we should automatically bind
+secret name along with the key.
+
+## Proposal
+
+The overall idea is to add more binding ways into `dataInputs` field.
+
+For k8s `secret`, we can add `toSecretKeyRef` field into `dataInputs`:
```
zzxwill, line 122:
```suggestion
For k8s `secret`, we can add `toSecretKeyRef` field into `dataInputs` for component `webapp`:
```

zzxwill, line 122: At the first glance, I think it is for workload `db` and find it doesn't make sense. And then I find it should be appended for workload `webapp`. So just for clarification, I made the modification:)

design/auto-binding.md
```diff
@@ -0,0 +1,199 @@
...
+
+```
+      dataInputs:
+        - valueFrom:
+            dataOutputName: alibaba-rdspostgresql-conn
+          toSecretKeyRef:
+            - secretKey: endpoint
+              path: spec.template.spec.containers[0].env[0].valueFrom
```
zzxwill, line 130: Use `fieldpath` instead?

zzxwill:
LGTM

design/auto-binding.md
wonderflow: good

design/auto-binding.md
wonderflow: This is an object field, I'm not sure if it's confusing to use the same name.

design/auto-binding.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+The oam runtime will automatically bind secret name and key to the workload.
+
+This means below fragment will be injected automatically.
+```yaml
+                  valueFrom:
+                    name: alibaba-rdspostgresql-conn
+                    secretKeyRef:
+                      key: endpoint
```
ryanzhang-oss, line 197: shouldn't this be
```yaml
           secretKeyRef:
                name: alibaba-rdspostgresql-conn
                key: endpoint
```
instead?

PR #51, opened 2020-06-27
Moving utils from addon local to runtime.

artursouza:


pkg/oam/util/test_utils.go
```diff
@@ -0,0 +1,90 @@
+package util
```
wonderflow, line 1: we can move it to separate package outside of `pkg` to increase the test coverage

artursouza: I am adding tests right now. See latest commit.

PR #52, opened 2020-06-27
[Component Versioning] fix saving right status of Component into  ControllerRevision

captainroy-hy:
**Problem description**

During creating ControllerRevision, the `component.status.latestRevision` being saved into `controllerRevision.data` is from the old component but not the new one. It makes confusing `controllerRevision.data`  that contains the `component.spec` of the current component but `component.status.latestRevision` of the previous one .

**Solution**

Update `component.status` before constructing and creating ControllerRevision.


wonderflow:
Good catch, can you write a test for this case?

captainroy-hy:
> Good catch, can you write a test for this case?

Yes, I'm still working on it.

PR #54, opened 2020-06-29
change dataInput.Matchers -> dataOutput.Conditions

wonderflow:


apis/core/v1alpha2/core_types.go
```diff
@@ -384,6 +386,12 @@ type DataOutput struct {
 
 	// FieldPath refers to the value of an object's field.
 	FieldPath string `json:"fieldPath,omitempty"`
+
+	// Matchers specify the requirements to match a value.
+	// Different requirements are AND-ed together.
+	// If no matcher is specified, it is by default to check not empty.
+	// +optional
+	Matchers []DataMatcherRequirement `json:"matchers,omitempty"`
```
hongchaodeng, line 394: If we put this into DataOutput, this doesn't mean "Matchers" anymore.
It is "Conditions":

```
conditions:
- fieldPath: status.state
  op: eq
  value: Running
```

hongchaodeng: And the comment needs to be changed:

```Go
// Conditions specify the conditions that should be satisfied before emitting a data output.
Conditions []ConditionRequirement
```

hongchaodeng:
The idea and code LGTM.
I think the name should change a bit.

hongchaodeng:
lgtm

examples/dependency/README.md
```diff
@@ -1,15 +1,173 @@
...
+kind: ApplicationConfiguration
+metadata:
+  name: example-appconfig
+spec:
+  components:
+    - componentName: source
+      dataOutputs:
+        - name: example-key
+          fieldPath: "spec.key"
+          matchers:
```
zzxwill, line 112:
```suggestion
          conditions:
```

zzxwill: Please also replace `matchers` in files and filenames.

apis/core/v1alpha2/core_types.go
zzxwill: Putting `Conditions` in `dataOutputs` parts seems confusing to me. If the condition turns out to be `false`, I might mistakenly think the outer controller will prevent the deployment of the component. 

Besides, if we set two conditions C1 and C2 for component A, and component B and C depends on component A, there will be confusion when B and C can move on.
By keeping the original design, ie C1 and C2 set in component B and C respectively, if C1 == true, component B can go on; if C2 == false, component C will hold on. This situation seems more efficient. But I admit I could not make a real example for this situation.

examples/dependency/README.md
```diff
@@ -1,15 +1,173 @@
...
+```shell script
+$ kubectl get foo.example.com source -o yaml
+apiVersion: example.com/v1
+kind: Foo
+metadata:
+  name: source
+  namespace: default
+```
+
+And the workload of `sink` component also don't have `spec.key`:
```
ryanzhang-oss, line 52: the sink should not exist!

pkg/dependency/dag.go
```diff
@@ -28,6 +28,8 @@ type SinksPerSource struct {
 type Source struct {
 	// ObjectRef refers to the object this source come from.
 	ObjectRef *corev1.ObjectReference
+
+	Matchers []v1alpha2.ConditionRequirement
```
ryanzhang-oss, line 32: Would you consider not using ObjectReference and just spell out all the needed fields?  Not sure why we need to borrow an object reference

examples/dependency/README.md
wonderflow: yes, I'll check it

pkg/dependency/dag.go
wonderflow: Why don't you want borrow object reference here? They're exactly same

PR #56, opened 2020-06-30
clean up previous PRs and add the workloadRef check in appConfig

ryanzhang-oss:
clean up previous PRs, fix tests and add the workloadRef check in the appConfig controller

### DIFF
--- a/pkg/controller/oam/applicationconfiguration/apply.go
+++ b/pkg/controller/oam/applicationconfiguration/apply.go
@@ -19,39 +19,39 @@ package applicationconfiguration
 import (
 	"context"
 
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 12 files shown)


pkg/controller/oam/applicationconfiguration/apply.go
```diff
@@ -134,25 +141,9 @@ func findDereferencedScopes(statusScopes []v1alpha2.WorkloadScope, scopes []unst
...
 
-func (a *workloads) applyScope(ctx context.Context, wl Workload, s unstructured.Unstructured) error {
-	workloadRef := runtimev1alpha1.TypedReference{
-		APIVersion: wl.Workload.GetAPIVersion(),
-		Kind:       wl.Workload.GetKind(),
-		Name:       wl.Workload.GetName(),
-	}
-
-	// Get Scope again to make sure we work with the most up-to date object.
-	scopeObject := unstructured.Unstructured{}
```
artursouza, line 145: I looked back in the code and realized this is not needed anymore. Thanks for removing it.

pkg/controller/oam/applicationconfiguration/apply.go
```diff
@@ -19,73 +19,80 @@ package applicationconfiguration
...
 
 type workloads struct {
 	client    resource.Applicator
 	rawClient client.Client
 }
 
-func (a *workloads) Apply(ctx context.Context, namespace string, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
+func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
+	// they are all in the same namespace
+	var namespace = w[0].Workload.GetNamespace()
```
wonderflow, line 64: what  if workload slice is empty?

ryanzhang-oss: Is that even legal? I am assuming that we will have an admission hook

PR #74, opened 2020-07-02
refactor dependency engine to reuse and fit in reconcile loop

hongchaodeng:
Per discussion with @wonderflow , the previous DAGManager loop has unexpected corner cases like handling orphan-ed AppConfigs and dealing with Workload's traits and scopes.

A better implementation solution would be to base on existing reconcile loop and find the dependencies in each reconcile. If dependency is not satisfied, it would requeue and retry later.


pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -81,24 +82,34 @@ type components struct {
...
+	res := make([]Workload, 0, len(ac.Spec.Components))
+	for i, acc := range ac.Spec.Components {
+		ready, err := r.handleDependency(ctx, workloads[i], acc, dag)
+		if err != nil {
+			return nil, false, err
 		}
-		if w == nil { // depends on other resources. Not creating now.
+		if !ready { // Will not render this workload if dependency is not satisfied. Requeue and retry later.
+			hasDep = true
 			continue
```
wonderflow, line 107: I suggest we don't continue here, but give a flag(dependency not ready), workload with this flag will not be applied, but status and scope can still be updated.

In this case, AppConfig.status will show what happening for this component instead of show nothing.

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+		return false, errors.Wrap(err, fmt.Sprintf("handleDataInput for workload (%s/%s) failed", w.Workload.GetNamespace(), w.Workload.GetName()))
+	}
+	if !ready {
+		return false, nil
+	}
+
+	for i, ct := range acc.Traits {
+		ready, err := r.handleDataInput(ctx, ct.DataInputs, dag, &w.Traits[i])
+		if err != nil {
+			return false, errors.Wrap(err, fmt.Sprintf("handleDataInput for trait (%s/%s) failed", w.Traits[i].GetNamespace(), w.Traits[i].GetName()))
```
wonderflow, line 405:
```suggestion
			return false, errors.Wrap(err, fmt.Sprintf("handleDataInput for trait (%s/%s) failed", ct.GetNamespace(), ct.GetName()))
```

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+	}
+	return true, nil
+}
+
+func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInput, dag *dependency.DAG, obj *unstructured.Unstructured) (bool, error) {
 	for _, in := range ins {
-		dag.AddSink(in.ValueFrom.DataOutputName, obj, attaches, in.ToFieldPaths)
+		s, ok := dag.Sources[in.ValueFrom.DataOutputName]
+		if !ok {
+			return false, errors.New("DataOutput name doesn't exist")
```
wonderflow, line 418: we should indicate which name here

pkg/dependency/dag.go
```diff
@@ -32,61 +19,17 @@ type Source struct {
 	Matchers []v1alpha2.ConditionRequirement
```
wonderflow, line 19: Name it Conditions better.

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+		return val, false, err
+	}
+	if !ok {
+		return val, false, nil
+	}
+
+	return val, true, nil
+}
+
+func matchValue(ms []v1alpha2.ConditionRequirement, val string, paved *fieldpath.Paved) (bool, error) {
```
ryanzhang-oss, line 481: Please add a UT for this function

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+
+		err = fillValue(obj, in.ToFieldPaths, val)
+		if err != nil {
+			return false, errors.Wrap(err, "fillValue failed")
+		}
+	}
+	return true, nil
+}
+
+func fillValue(obj *unstructured.Unstructured, fs []string, val string) error {
```
ryanzhang-oss, line 436: Please add a UT for this function

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+			return false, errors.Wrap(err, fmt.Sprintf("handleDataInput for trait (%s/%s) failed", w.Traits[i].GetNamespace(), w.Traits[i].GetName()))
+		}
+		if !ready {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInput, dag *dependency.DAG, obj *unstructured.Unstructured) (bool, error) {
```
ryanzhang-oss, line 414: Please add a UT for this function

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
 	}
 }
 
-func addDataInputsToDAG(dag *dependency.DAG, ins []v1alpha2.DataInput, obj *unstructured.Unstructured, attaches []unstructured.Unstructured) {
+func (r *components) handleDependency(ctx context.Context, w *Workload, acc v1alpha2.ApplicationConfigurationComponent, dag *dependency.DAG) (bool, error) {
```
ryanzhang-oss, line 393: Please add a UT for this function

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+			if err != nil {
+				return false, err
+			}
+		} else {
+			checkVal = val
+		}
+
+		switch m.Operator {
+		case v1alpha2.ConditionEqual:
+			if m.Value != checkVal {
```
ryanzhang-oss, line 501: do we assume it's a primitive type like string or int?

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+				return false, nil
+			}
+		case v1alpha2.ConditionNotEqual:
+			if m.Value == checkVal {
```
ryanzhang-oss, line 505: do we assume it's a primitive type like string or int?

pkg/controller/oam/applicationconfiguration/render.go
```diff
@@ -392,8 +390,126 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+				return false, nil
+			}
+		case v1alpha2.ConditionNotEmpty:
+			if checkVal == "" {
```
ryanzhang-oss, line 509: do we assume it's string only?

pkg/controller/oam/applicationconfiguration/render.go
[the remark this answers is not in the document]
hongchaodeng: Currently there is no plan to support dependency w.r.t. scope. Only workloads and traits.
This is because I do not want to bloat the feature to cover scenarios that do not have real world use cases.

pkg/controller/oam/applicationconfiguration/render.go
hongchaodeng: Currently only string type is supported. This should fix all the use cases as mentioned in the design doc.
We don't see any use case for int yet.

hongchaodeng: For complex objects, we are going to provide more high level functions like AutoBinding to fill the gap.

hongchaodeng:
@wonderflow The AppConfig now will show the dependency status:

```
  status:
    dependency:
      unsatisfied:
      - from:
          apiVersion: example.com/v1
          kind: Foo
          name: source
          fieldPath: status.key
        to:
          apiVersion: example.com/v1
          kind: Foo
          name: sink
          fieldPaths:
          - spec.key
```

hongchaodeng:
The reason why previous implementation just take the whole Workload and its Traits when dependency not met is just to show the prototype.

Now the current implementation has changed that to mark each Workload and Trait whether they are "ready" to apply. This is a more robust solution and easier to debug along with dependency status.

pkg/controller/oam/applicationconfiguration/render.go
hongchaodeng: covered in TestDependency

hongchaodeng:
Added tests:


```
--- PASS: TestMatchValue (0.00s)
    --- PASS: TestMatchValue/No_conditions_with_nonempty_value_should_match (0.00s)
    --- PASS: TestMatchValue/eq_condition_with_different_value_should_not_match (0.00s)
    --- PASS: TestMatchValue/notEmpty_condition_with_nonempty_value_should_match (0.00s)
    --- PASS: TestMatchValue/eq_condition_with_different_value_from_FieldPath_should_not_match (0.00s)
    --- PASS: TestMatchValue/No_conditions_with_empty_value_should_not_match (0.00s)
    --- PASS: TestMatchValue/eq_condition_with_same_value_should_match (0.00s)
    --- PASS: TestMatchValue/notEq_condition_with_different_value_should_match (0.00s)
    --- PASS: TestMatchValue/notEq_condition_with_same_value_should_not_match (0.00s)
    --- PASS: TestMatchValue/notEmpty_condition_with_empty_value_should_not_match (0.00s)
    --- PASS: TestMatchValue/eq_condition_with_same_value_from_FieldPath_should_match (0.00s)

--- PASS: TestDependency (0.01s)
    --- PASS: TestDependency/DataOutputName_doesn't_exist (0.00s)
    --- PASS: TestDependency/Workload_depends_on_another_Workload_that's_ready (0.00s)
    --- PASS: TestDependency/Workload_depends_on_a_Trait_that's_ready (0.00s)
    --- PASS: TestDependency/Trait_depends_on_a_Workload_that's_ready (0.00s)
    --- PASS: TestDependency/Trait_depends_on_another_unreadyTrait_that's_unready (0.00s)
    --- PASS: TestDependency/Trait_depends_on_another_unreadyTrait_that's_ready (0.00s)
    --- PASS: TestDependency/Workload_depends_on_another_Workload_that's_unready (0.00s)
    --- PASS: TestDependency/Workload_depends_on_a_Trait_that's_unready (0.00s)
    --- PASS: TestDependency/Trait_depends_on_a_Workload_that's_unready (0.00s)
```

hongchaodeng:
Since the initial submission the following comments  has been addressed:

- [X] Add status to indicate the progress of dependency waiting
- [X] Handle unreadiness of Workload/Trait resources individually instead of bundling them all together.
- [X] Add UT to cover all major cases as shown above

@My-pleasure is trying to write e2e tests to cover dependency user story, which will be up for review early next week. I would like to merge this PR first instead of coupling his commits here.

examples/dependency/demo.yaml
```diff
@@ -8,6 +8,8 @@ spec:
     kind: Foo
     metadata:
       name: source
+    status:
+      key: test
```
wonderflow, line 12: This should not exist at initial time.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -247,13 +255,26 @@ type Workload struct {
...
 	Scopes []unstructured.Unstructured
 }
 
+// A Trait produced by an OAM ApplicationConfiguration.
+type Trait struct {
+	Object unstructured.Unstructured
+
+	// Unready indicates whether this resource is unready to apply
+	// because of dependency not ready
+	Unready bool
```
wonderflow, line 275: What if the outside Unready is true while the `Unready` field here is false? Does it mean workload will apply before all trait is ready?

pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -63,19 +63,26 @@ func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus,
 	// they are all in the same namespace
 	var namespace = w[0].Workload.GetNamespace()
 	for _, wl := range w {
-		if err := a.client.Apply(ctx, wl.Workload, ao...); err != nil {
-			return errors.Wrapf(err, errFmtApplyWorkload, wl.Workload.GetName())
+		if !wl.Unready {
+			err := a.client.Apply(ctx, wl.Workload, ao...)
```
wonderflow, line 67: I think we can't apply workload before its trait still unready. They should work as a whole.

wonderflow:
I think this PR will also fix #89, right? @hongchaodeng

examples/dependency/demo.yaml
hongchaodeng: Let me comment it out. Having this without typing every time is very convenient.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
hongchaodeng: This does not have any indication on ordering. It will comply with whatever ordering being applied but skip unready ones.

hongchaodeng:
> I think this PR also fixes #89, right? 

Yes. The inputs handling logic is AND-ed.

```Go
	uds := make([]v1alpha2.UnstaifiedDependency, 0)
	for _, in := range ins {
		s, ok := dag.Sources[in.ValueFrom.DataOutputName]
		val, ready, err := r.checkSourceReady(ctx, s)
		if !ready {
			uds = append(uds, makeUnsatisfiedDependency(obj, s, in))
			return uds, nil
		}
...
	}
```

pkg/controller/v1alpha2/applicationconfiguration/apply.go
hongchaodeng: Why do we assume such logic in dependency? What's the reasoning?

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -81,27 +89,34 @@ type components struct {
...
+
+		workloads = append(workloads, w)
 	}
 
-	if !dag.IsEmpty() {
-		dependency.GlobalManager.AddDAG(ac.GetNamespace()+"/"+ac.GetName(), dag)
+	ds := &v1alpha2.DependencyStatus{}
+	res := make([]Workload, 0, len(ac.Spec.Components))
+	for i, acc := range ac.Spec.Components {
+		unsatsified, err := r.handleDependency(ctx, workloads[i], acc, dag)
```
ryanzhang-oss, line 108: nit: spell
```suggestion
		unsatisified, err := r.handleDependency(ctx, workloads[i], acc, dag)
```

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -421,8 +419,154 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInput, dag *dag, obj *unstructured.Unstructured) ([]v1alpha2.UnstaifiedDependency, error) {
+	uds := make([]v1alpha2.UnstaifiedDependency, 0)
 	for _, in := range ins {
-		dag.AddSink(in.ValueFrom.DataOutputName, obj, attaches, in.ToFieldPaths)
+		s, ok := dag.Sources[in.ValueFrom.DataOutputName]
+		if !ok {
+			return nil, errors.Wrapf(ErrDataOutputNotExist, "DataOutputName (%s)", in.ValueFrom.DataOutputName)
+		}
+		val, ready, err := r.checkSourceReady(ctx, s)
+		if err != nil {
```
ryanzhang-oss, line 477: so this means that a "ready" source can be fliped to be "unready" due to network or some other errors. I think we at least should put a TODO here to fine-tune that user experience in the next PR.

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -421,8 +419,154 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+	for _, fp := range fs {
+		err := paved.SetString(fp, val)
+		if err != nil {
+			return fmt.Errorf("paved.SetString() failed: %w", err)
+		}
+	}
+	return nil
+}
+
+func (r *components) checkSourceReady(ctx context.Context, s *dagSource) (string, bool, error) {
```
ryanzhang-oss, line 504: I don't seem to see a Unit test here, maybe we can ask the student to add it?

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -421,8 +419,154 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
...
+				APIVersion: obj.GetAPIVersion(),
+				Kind:       obj.GetKind(),
+				Name:       obj.GetName(),
+			},
+			FieldPaths: in.ToFieldPaths,
+		},
+	}
+}
+
+func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInput, dag *dag, obj *unstructured.Unstructured) ([]v1alpha2.UnstaifiedDependency, error) {
```
ryanzhang-oss, line 469: I also don't seem to see a unit test here

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -421,8 +419,154 @@ func addDataOutputsToDAG(dag *dependency.DAG, outs []v1alpha2.DataOutput, obj *u
 	}
 }
 
-func addDataInputsToDAG(dag *dependency.DAG, ins []v1alpha2.DataInput, obj *unstructured.Unstructured, attaches []unstructured.Unstructured) {
+func (r *components) handleDependency(ctx context.Context, w *Workload, acc v1alpha2.ApplicationConfigurationComponent, dag *dag) ([]v1alpha2.UnstaifiedDependency, error) {
```
ryanzhang-oss, line 422: and this one

pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -63,19 +63,26 @@ func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus,
 	// they are all in the same namespace
 	var namespace = w[0].Workload.GetNamespace()
 	for _, wl := range w {
-		if err := a.client.Apply(ctx, wl.Workload, ao...); err != nil {
-			return errors.Wrapf(err, errFmtApplyWorkload, wl.Workload.GetName())
+		if !wl.Unready {
```
ryanzhang-oss, line 66: Does this new case have unit test?

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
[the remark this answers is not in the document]
hongchaodeng: The reason is that the default of bool is false.
The default of any Workload/Trait is actually ready (= !unready). The dataInputs field might not exist so it might not go into the dependency logic.

What about change the name to `HasDep`?

apis/core/v1alpha2/core_types.go
```diff
@@ -357,10 +352,37 @@ type WorkloadStatus struct {
...
+// one pair of objects.
+type UnstaifiedDependency struct {
+	From DependencyFromObject `json:"from"`
+	To   DependencyToObject   `json:"to"`
+}
+
+// DependencyFromObject represents the object that dependency data comes from.
+type DependencyFromObject struct {
+	runtimev1alpha1.TypedReference `json:",inline"`
+	FieldPath                      string `json:"fieldPath,omitempty"`
```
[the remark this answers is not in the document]
hongchaodeng: It is allowed.
The status will show each dependency for the same object in individual item:

```
status:
  dependencies:
  - from:
      name: source-a
    to:
      name: sink-a
  - from:
      name: source-b
    to:
      name: sink-a
```

But actually in current implementation, the runtime will save computing power and not print the second one as long as it finds any one dependency for a sink not satisfied.

pkg/controller/v1alpha2/applicationconfiguration/render.go
hongchaodeng: This is the reason we put the logic into reconcile and let controller runtime deal with error handling.
If such transient error happens, the AppConfig status will show it and from users' point of view they should be able to see it and wait for the network hiccup to pass away.

pkg/controller/v1alpha2/applicationconfiguration/render.go
hongchaodeng: Let me answer all questions about the UT here.
All of the mentioned methods are covered in TestDependency. I have used `go tool cover` to check that.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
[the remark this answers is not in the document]
hongchaodeng: I have changed this to HasDep. I think having a default false value is still very important in preventing mistaken init place.

My-pleasure:
LGTM, I have verified this and will follow up with a e2e test PR shortly.

PR #85, opened 2020-07-06
fix(trait):make the default trait name not the same as the component name

### LINKED ISSUE #66: Make the default trait name not the same as the component name
Currently, we use the component name as the trait name in the `setTraitProperties` function. It will be good for the name to indicate that it belongs to a `trait`

Ghostbaby:
Make the default trait name not the same as the component name

Fixes #66

Signed-off-by: zhuhuijun <zhuhuijunzhj@gmail.com>


wonderflow:
in Kubernetes, children resources have the name of their parents as prefix like:

* Deployment: deploymentname
* ReplicaSet: deploymentname-dsjn3ndshh
* Pod: deploymentname-dsjn3ndshh-idsjn

I suggest we do something like it here.

* ComponentName: my-comp
* TraitName: my-comp-trait-dsjn3ndshh

pkg/oam/util/helper.go
```diff
@@ -32,6 +32,11 @@ var (
 	ReconcileWaitResult = reconcile.Result{RequeueAfter: 30 * time.Second}
 )
 
+const (
+	//TraitPrefixKey is prefix of trait name
+	TraitPrefixKey = "trait"
```
ryanzhang-oss, line 37: I don't seem to see this being used anywhere else, why make it public?

pkg/oam/util/helper.go
```diff
@@ -197,3 +202,8 @@ func Object2Map(obj interface{}) (map[string]interface{}, error) {
 	err = json.Unmarshal(bts, &res)
 	return res, err
 }
+
+// GenTraitName generate trait name
```
ryanzhang-oss, line 206: please add a unit test for this function

pkg/oam/util/helper.go
```diff
@@ -197,3 +202,8 @@ func Object2Map(obj interface{}) (map[string]interface{}, error) {
...
+func GenTraitName(componentName string) string {
+	return fmt.Sprintf("%s-%s", TraitPrefixKey, componentName)
```
ryanzhang-oss, line 208: I think @wonderflow has a good suggestion on the naming convention

pkg/oam/util/helper.go
Ghostbaby: ok

pkg/oam/util/helper.go
Ghostbaby: It will be used in `GenTraitName`.

pkg/oam/util/helper.go
Ghostbaby: ok

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -165,12 +165,13 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 }
 
 func (r *components) renderTrait(ctx context.Context, ct v1alpha2.ComponentTrait, namespace, componentName string, ref *metav1.OwnerReference, dag *dependency.DAG) (*unstructured.Unstructured, *v1alpha2.TraitDefinition, error) {
+	traitName := util.GenTraitName(componentName, ct.DeepCopy(), nil)
```
wonderflow, line 168: I think we must check if there's already an exist trait, before generate a new traitname

pkg/oam/util/helper.go
ryanzhang-oss: GenTraitName is in the same package....

pkg/oam/util/helper.go
```diff
@@ -197,3 +208,39 @@ func Object2Map(obj interface{}) (map[string]interface{}, error) {
...
+// a collisionCount to avoid hash collision. The hash will be safe encoded to
+// avoid bad words.
+func ComputeHash(trait *v1alpha2.ComponentTrait, collisionCount *int32) string {
+	componentTraitHasher := fnv.New32a()
+	DeepHashObject(componentTraitHasher, *trait)
+
+	// Add collisionCount in the hash if it exists.
+	if collisionCount != nil {
+		collisionCountBytes := make([]byte, 8)
+		binary.LittleEndian.PutUint32(collisionCountBytes, uint32(*collisionCount))
```
ryanzhang-oss, line 227: collisionCount is always passed in as new(int32) which I assume is a pointer to zero, I wonder why do we need this parameter?

pkg/controller/v1alpha2/applicationconfiguration/render.go
Ghostbaby: done

pkg/oam/util/helper.go
Ghostbaby: yes, I will remove this parameter.

zzxwill:
> in Kubernetes, children resources have the name of their parents as prefix like:
> 
> * Deployment: `deploymentname`
> * ReplicaSet: `deploymentname-dsjn3ndshh`
> * Pod: `deploymentname-dsjn3ndshh-idsjn`
> 
> I suggest we do something like it here.
> 
> * AppConfig: `appconfigname` // unique globally
> * Component: `componentname`  // unique globally
> * Trait: `componentname-trait-dsjn3ndshh` // generate a hash suffix
> 
> If we don't assign name, we can say there's only one instance for one kind of trait. So we can get from ownerRef before create new.

`unique globally` means cross namespaces? Or it will allow same names in different names?

pkg/oam/util/helper_test.go
```diff
@@ -292,3 +292,41 @@ var _ = Describe("Test unstructured related helper utils", func() {
...
+			},
+		}
+		for _, test := range test {
+			name := fmt.Sprintf("%s-%s-%s", test.name, util.TraitPrefixKey, util.ComputeHash(test.template))
+			traitName := util.GenTraitName(test.name, test.template)
+			Expect(traitName).Should(Equal(name))
+
+			new := test.template.DeepCopy()
+			new.Trait = runtime.RawExtension{
+				Object: &mts,
```
ryanzhang-oss, line 327: what does this test?

pkg/oam/util/helper_test.go
```diff
@@ -292,3 +292,41 @@ var _ = Describe("Test unstructured related helper utils", func() {
...
+			Spec: v1alpha2.ManualScalerTraitSpec{
+				ReplicaCount: 3,
+			},
+		}
+
+		test := []struct {
+			name                string
+			template            *v1alpha2.ComponentTrait
+			collisionCount      *int32
+			otherCollisionCount *int32
```
ryanzhang-oss, line 313: I don't seem to see those being used

pkg/oam/util/helper_test.go
```diff
@@ -292,3 +292,41 @@ var _ = Describe("Test unstructured related helper utils", func() {
 		}
 	})
 })
+
+var _ = Describe("Test GenTraitName helper utils", func() {
+	It("Test generate trait name", func() {
```
ryanzhang-oss, line 297: This test actually relies on the ComputeHash function. I wonder if we can have an "expected" value that is not computed by the ComputeHash function?

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -397,3 +403,39 @@ func addDataInputsToDAG(dag *dependency.DAG, ins []v1alpha2.DataInput, obj *unst
...
+
+		for _, cr := range crs.Items {
+			for _, owner := range cr.GetOwnerReferences() {
+				if owner.UID == ref.UID {
+					traitName = cr.GetName()
+				}
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
Ghostbaby: Maybe we can store traitName to annotations.

pkg/oam/util/helper_test.go
Ghostbaby: I will remove these parameters.

pkg/oam/util/helper_test.go
Ghostbaby: The main test is to pass in the same `ComponentTrait` to get the same hash value.

pkg/oam/util/helper_test.go
Ghostbaby: done

pkg/oam/util/helper_test.go
```diff
@@ -292,3 +292,46 @@ var _ = Describe("Test unstructured related helper utils", func() {
...
+			newTemplate := test.template.DeepCopy()
+			newTemplate.Trait = runtime.RawExtension{
+				Object: &mts,
+			}
+			newTraitName := util.GenTraitName(test.name, newTemplate)
+			Expect(oldTraitName).ShouldNot(Equal(newTraitName))
+
+			newTemplate.Trait = runtime.RawExtension{}
+
+			lastedTraitName := util.GenTraitName(test.name, newTemplate)
```
ryanzhang-oss, line 332: nit: lastTraitName ?

ryanzhang-oss, line 332: a table style test should not have logics in the last loop, instead, they should be test cases. You can move those to test cases and add a result (should equal or not equal) there.

pkg/controller/v1alpha2/applicationconfiguration/render.go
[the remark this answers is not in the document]
ryanzhang-oss: I think the appConfig CR actually has a status field that you can store the trait name

pkg/oam/util/helper_test.go
Ghostbaby: done

pkg/controller/v1alpha2/applicationconfiguration/render.go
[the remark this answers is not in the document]
Ghostbaby: ok, I will get the history trait name form the status field.

Ghostbaby:
@ryanzhang-oss ok, I will get the history trait name form the status field.

pkg/controller/v1alpha2/applicationconfiguration/render_test.go
```diff
@@ -907,6 +912,79 @@ func TestGetComponent(t *testing.T) {
 	}, c)
 }
 
+var scheme = runtime.NewScheme()
+
+func TestRenderTraitName(t *testing.T) {
```
ryanzhang-oss, line 917: IIRC, this is the case that a trait name exist. It seems that it misses the case that trait doesn't exist

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -426,3 +429,33 @@ func addDataInputsToDAG(dag *dependency.DAG, ins []v1alpha2.DataInput, obj *unst
 		dag.AddSink(in.ValueFrom.DataOutputName, obj, attaches, in.ToFieldPaths)
 	}
 }
+
+// GetTraitName return trait name
+func getTraitName(ac *v1alpha2.ApplicationConfiguration, componentName string,
+	ct *v1alpha2.ComponentTrait, t *unstructured.Unstructured) string {
+	var (
```
wonderflow, line 436: I assume we can check `t.GetName()` first, if this is not null, just return it. That's also the traitName specified by user.

pkg/controller/v1alpha2/applicationconfiguration/render_test.go
Ghostbaby: done

pkg/controller/v1alpha2/applicationconfiguration/render.go
Ghostbaby: done

Ghostbaby:
@wonderflow @zzxwill @ryanzhang-oss @sunny0826  Please help me review this PR. At the same time, the CI UT coverage calculation seems to have problems.

ryanzhang-oss:
@Ghostbaby  please kick the build process again, it seems to be stuck

PR #86, opened 2020-07-06
pass through labels and annotation from app-config to workload  or trait

### LINKED ISSUE #77: app-config pass labels and annotation to workload and trait
Internally, there are scenarios where we need to get corresponding Annotation information in the Controller of the Workload and trait, so we can do some business processing based on the annotations, such as namespace isolation
Workload and trait only processing those workload and traits where the information in the Annotation meets the conditions

If app-config can pass this information, it's very convenient, You can write generic annotations to Workload and traits without having to write them intentionally, In particular, writing workload annotations in app-config is a little more complicated

allenhaozi:
pass through app-config's labels and annotation  to its workload and trait

fixes #77

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render.go
@@ -119,6 +119,9 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 		return nil, errors.Wrapf(err, errFmtRenderWorkload, acc.ComponentName)
 	}
 
+	// pass through labels and annotation from app-config to workload
+	r.passThroughObjMeta(ac.ObjectMeta, w)
+
 	ref := metav1.NewControllerRef(ac, v1alpha2.ApplicationConfigurationGroupVersionKind)
 	w.SetOwnerReferences([]metav1.OwnerReference{*ref})
 	w.SetNamespace(ac.GetNamespace())
@@ -134,6 +137,9 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 		if t == nil { // Depends on other resources. Not creating it now.
 			continue
 		}
+
+		// pass through labels and annotation from app-config to trait
+		r.passThroughObjMeta(ac.ObjectMeta, t)
 		traits = append(traits, *t)
 		traitDefs = append(traitDefs, *traitDef)
 	}
@@ -264,6 +270,29 @@ func (r *components) getComponent(ctx context.Context, acc v1alpha2.ApplicationC
 		revisionName = c.Status.LatestRevision.Name
 	}
 	return c, revisionName, nil
+}
+
+// pass through labels and annotation from app-config to workload  or trait
+func (r *components) passThroughObjMeta(oMeta metav1.ObjectMeta, u *unstructured.Unstructured) {
+	mergeMap := func(src, dst map[string]string) map[string]string {
+		if len(src) == 0 {
+			return dst
+		}
+		// make sure dst is initialized
+		if dst == nil {
+			dst = map[string]string{}
+		}
+		for k, v := range src {
+			if _, exist := dst[k]; !exist {
+				dst[k] = v
+			}
+		}
+		return dst
+	}
+	// pass app-config labels
+	u.SetLabels(mergeMap(oMeta.GetLabels(), u.GetLabels()))
+	// pass app-config annotation
+	u.SetAnnotations(mergeMap(oMeta.GetAnnotations(), u.GetAnnotations()))
 }
 
 // A ResourceRenderer renders a Kubernetes-compliant YAML resource into an
--- a/pkg/controller/v1alpha2/applicationconfiguration/render_test.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render_test.go
@@ -18,6 +18,7 @@ package applicationconfiguration
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	"github.com/crossplane/crossplane-runtime/pkg/test"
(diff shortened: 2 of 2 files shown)


pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -266,6 +273,34 @@ func (r *components) getComponent(ctx context.Context, acc v1alpha2.ApplicationC
...
+func (r *components) passThroughObjMeta(oMeta metav1.ObjectMeta,u *unstructured.Unstructured) error{
+	// pass app-config labels
+	curLabels := oMeta.GetLabels()
+	if curLabels != nil && len(curLabels) > 0 {
+		// workload or trait labels
+		dstLabels := u.GetLabels()
+		if dstLabels == nil {
+			dstLabels = map[string]string{}
+		}
+		mergo.Map(&dstLabels,curLabels)
```
ryanzhang-oss, line 286: what does this do? We kinda out-source the entire core logic of this PR to this library.

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -266,6 +273,34 @@ func (r *components) getComponent(ctx context.Context, acc v1alpha2.ApplicationC
...
 }
 
+// pass through labels and annotation from app-config to workload  or trait
+func (r *components) passThroughObjMeta(oMeta metav1.ObjectMeta,u *unstructured.Unstructured) error{
+	// pass app-config labels
+	curLabels := oMeta.GetLabels()
+	if curLabels != nil && len(curLabels) > 0 {
+		// workload or trait labels
+		dstLabels := u.GetLabels()
+		if dstLabels == nil {
```
ryanzhang-oss, line 283: I don't think this can be nil

wonderflow:
please sign-off by `git commit -s --amend`
and configure your IDE to go format your code.

pkg/controller/v1alpha2/applicationconfiguration/render.go
allenhaozi: Merging maps is a simple logic, encapsulates a function to replace external library dependencies

ryanzhang-oss:
please fix the build failure with "make check-diff" on your local developing environment

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -266,6 +272,36 @@ func (r *components) getComponent(ctx context.Context, acc v1alpha2.ApplicationC
 	return c, revisionName, nil
 }
 
+// pass through labels and annotation from app-config to workload  or trait
+func (r *components) passThroughObjMeta(oMeta metav1.ObjectMeta, u *unstructured.Unstructured) {
+	mergeFn := func(src map[string]string, dst *map[string]string) {
```
wonderflow, line 277: Why use pointer for dst? In fact, variable of map[string]string is already a pointer

allenhaozi: go maps is not complete references. it only passed by reference once initialized
To avoid panic,  check the dst whether it is initialized or not, if not mergeFn will initialize it
```
package main

import "fmt"

func fn(m map[int]int) {
	m = make(map[int]int)
}

func main() {
	var m map[int]int
	fn(m)
	fmt.Println(m == nil) // got true, hope false
}
```

wonderflow: I see, can you help add a unit case that the dst will be nil?

allenhaozi:
> Thank you very much for working on this



> please fix the build failure with "make check-diff" on your local developing environment

thank you for your help

pkg/controller/v1alpha2/applicationconfiguration/render.go
allenhaozi: this unit covers the scenario where dst is empty, it's has been added
```
	t.Run("workload and trait have no labels and annotation", func(t *testing.T) {
		var u unstructured.Unstructured
		c.passThroughObjMeta(ac.ObjectMeta, &u)
		got := u.GetLabels()
        ...
```

wonderflow: Usually we return the result and reduce variables:
```

mergeFn  := func(src map[string]string, dst map[string]string) map[string]string {
}

u.SetLabels(mergeFn(oMeta.GetAnnotations(), u.GetAnnotations()))
```

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -274,32 +274,25 @@ func (r *components) getComponent(ctx context.Context, acc v1alpha2.ApplicationC
 
 // pass through labels and annotation from app-config to workload  or trait
 func (r *components) passThroughObjMeta(oMeta metav1.ObjectMeta, u *unstructured.Unstructured) {
-	mergeFn := func(src map[string]string, dst *map[string]string) {
+	mergeFn := func(src map[string]string, dst map[string]string) map[string]string {
 		if len(src) > 0 {
```
wonderflow, line 278: we won't need to check it as `for .. range ` can cover this case

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -266,6 +272,29 @@ func (r *components) getComponent(ctx context.Context, acc v1alpha2.ApplicationC
 	return c, revisionName, nil
 }
 
+// pass through labels and annotation from app-config to workload  or trait
+func (r *components) passThroughObjMeta(oMeta metav1.ObjectMeta, u *unstructured.Unstructured) {
+	mergeFn := func(src map[string]string, dst map[string]string) map[string]string {
```
wonderflow, line 277:
```suggestion
	mergeMap := func(src map[string]string, dst map[string]string) map[string]string {
```

PR #90, opened 2020-07-08
Add repo machinery to allow oam-kubernetes-runtime to publish artifacts using Jenkins

hasheddan:
This updates the repository machinery to make it compatible with publishing artifacts using the Crossplane Jenkins instance. Draft until the proper credentials are set up on the Jenkins side, and we determine exactly where these artifacts should be published / what they should be named.


Makefile
```diff
@@ -13,12 +13,20 @@ PLATFORMS ?= linux_amd64 linux_arm64
...
+# S3_BUCKET ?= crossplane.releases
+-include build/makelib/output.mk
+
+# ====================================================================================
+# Setup Helm
+
+HELM_BASE_URL = https://charts.crossplane.io
+HELM_S3_BUCKET = crossplane.charts
+HELM_CHARTS_DIR=$(ROOT_DIR)/charts
+HELM_CHARTS = oam-core-runtime
```
hasheddan, line 27: There are some inconsistencies with naming here. The helm chart is named `oam-core-runtime`. The binary and Docker image are `oam-runtime`. The repo is `oam-kubernetes-runtime`. Any thoughts on a consistent name?

Makefile
```diff
@@ -13,12 +13,20 @@ PLATFORMS ?= linux_amd64 linux_arm64
...
 
 # ====================================================================================
-# Setup Images
+# Setup Output
 
-# even though this repo doesn't build images (note the no-op img.build target below),
-# some of the init is needed for the cross build container, e.g. setting BUILD_REGISTRY
--include build/makelib/image.mk
-img.build:
+# S3_BUCKET ?= crossplane.releases
```
hasheddan, line 18: @jbw976 wasn't sure if we wanted to push these to the S3 bucket also? We do this for Crossplane but not for providers. Not sure if the helm chart is the differentiator.

Jenkinsfile
```diff
@@ -97,6 +97,25 @@ pipeline {
...
+    //             sh "./build/run make -j\$(nproc) publish BRANCH_NAME=${BRANCH_NAME} AWS_ACCESS_KEY_ID=${AWS_USR} AWS_SECRET_ACCESS_KEY=${AWS_PSW} GIT_API_TOKEN=${GITHUB_UPBOUND_BOT}"
+    //             script {
+    //                 if (BRANCH_NAME == 'master') {
+    //                     lock('promote-job') {
+    //                         sh "./build/run make -j\$(nproc) promote BRANCH_NAME=master CHANNEL=master AWS_ACCESS_KEY_ID=${AWS_USR} AWS_SECRET_ACCESS_KEY=${AWS_PSW}"
+    //                     }
+    //                 }
+    //             }
+    //         }
+    //     }
```
ryanzhang-oss, line 118: why are those commented out?

Makefile
ryanzhang-oss: is this where the image lives? If so, can anyone push an image there?

Makefile
ryanzhang-oss: I have no particular opinion on the specific names but I agree that it's better to be the same. How about they are all called oam-core-runtime ? Where is the image name located?

Makefile
```diff
@@ -13,12 +13,20 @@ PLATFORMS ?= linux_amd64 linux_arm64
...
+-include build/makelib/output.mk
+
+# ====================================================================================
+# Setup Helm
+
+HELM_BASE_URL = https://charts.crossplane.io
```
ryanzhang-oss, line 24: I wonder who can push helm charts there?

Makefile
```diff
@@ -34,11 +42,32 @@ GO_TEST_PARALLEL := $(shell echo $$(( $(NPROCS) / 2 )))
...
 -include build/makelib/golang.mk
 
+# ====================================================================================
+# Setup Images
+# Due to the way that the shared build logic works, images should
+# all be in folders at the same level (no additional levels of nesting).
+
+DOCKER_REGISTRY = crossplane
+IMAGE_DIR=$(ROOT_DIR)
+IMAGE = $(BUILD_REGISTRY)/oam-runtime-$(ARCH)
```
ryanzhang-oss, line 58: looks like the image name also contains (ARCH), I guess it's not very important in the sense that it's not user facing

charts/oam-core-runtime/.gitignore
```diff
@@ -0,0 +1,2 @@
+values.yaml
```
ryanzhang-oss, line 1: I didn't get why the `value.yaml` file is renamed and also ignored here. What is that `values.yaml-e` file?

I also don't know why it is renamed.

Makefile
```diff
@@ -34,11 +42,32 @@ GO_TEST_PARALLEL := $(shell echo $$(( $(NPROCS) / 2 )))
...
+GO_SUBDIRS += cmd pkg apis
 GO111MODULE = on
 -include build/makelib/golang.mk
 
+# ====================================================================================
+# Setup Images
+# Due to the way that the shared build logic works, images should
+# all be in folders at the same level (no additional levels of nesting).
+
+DOCKER_REGISTRY = crossplane
```
ryanzhang-oss, line 56: Who can push images there?

Makefile
negz: I believe this is where the Helm chart lives. I presume the image would live at https://hub.docker.com/u/crossplane.

Makefile
wonderflow: I think they should all be called `oam-kubernetes-runtime` align with the repo.

charts/oam-core-runtime/.gitignore
hasheddan: `values.yaml` is now `values.yaml.tmpl`. This is just how the `build` machinery works. It uses `values.yaml.tmpl` to generate `values.yaml` during the build. This allows you to do something like this https://github.com/crossplane/oam-kubernetes-runtime/pull/90/files#diff-b67911656ef5d18c4ae36cb6741b7965R28

Makefile
hasheddan: This is handled by Jenkins, which has the credentials to push the chart. You can see more here https://crossplane.io/docs/v0.12/release/release-process

Makefile
hasheddan: @ryanzhang-oss image name is set here https://github.com/crossplane/oam-kubernetes-runtime/pull/90/files#diff-b67911656ef5d18c4ae36cb6741b7965R58

Does `oam-kubernetes-runtime` sound good to you?

Makefile
hasheddan: Handled by Jenkins just like the helm chart :)

Makefile
hasheddan: We will push multi-arch images, see https://hub.docker.com/r/crossplane/crossplane/tags for example of how each image has an AMD and an ARM flavor

Makefile
hasheddan: I think the helm chart actually lives at `crossplane.charts` (https://github.com/crossplane/oam-kubernetes-runtime/pull/90/files#diff-b67911656ef5d18c4ae36cb6741b7965R25), so I am going to wait for @jbw976 to confirm the purpose of this bucket 👍

jbw976: @hasheddan we do not want to push anything from this repo to a "releases" s3 bucket. We accidentally copy/paste did this for the providers when we bootstrapped them, then realized they were overwriting previous crossplane releases 🤯, see https://github.com/crossplane/crossplane/issues/1320 for more details on that mistake.

Publishing the image to dockerhub and the helm chart to crossplane.charts should be all we need for this repo :)

Makefile
ryanzhang-oss: sure

ryanzhang-oss:
Is the PR ready to merge?

hasheddan:
@ryanzhang-oss I think this is ready now :+1:

Makefile
```diff
@@ -125,23 +123,16 @@ oam-kubernetes-runtime.help:
...
-
-# Push the docker image
-docker-push:
-	docker push ${IMG}
+.PHONY: oam-kubernetes-runtime.help help-special kind-load e2e-setup e2e-test e2e-cleanup
 
 # load docker image to the kind cluster
 kind-load:
-	kind load docker-image $(IMG) || { echo >&2 "kind not installed or error loading image: $(IMG)"; exit 1; }
+	docker tag $(IMAGE) crossplane/oam-kubernetes-runtime:$(VERSION)
```
ryanzhang-oss, line 130: Can you point me to where/how is $(IMAGE) build?

Makefile
```diff
@@ -125,23 +123,16 @@ oam-kubernetes-runtime.help:
...
+	kind load docker-image crossplane/oam-kubernetes-runtime:$(VERSION) || { echo >&2 "kind not installed or error loading image: $(IMAGE)"; exit 1; }
 
-e2e-setup: docker-build kind-load
+e2e-setup: build kind-load
 	kubectl create namespace oam-system
-	helm install e2e ./charts/oam-kubernetes-runtime/ -n oam-system --set image.repository=$(IMG),image.tag='',image.pullPolicy='Never' --wait \
+	helm install e2e ./charts/oam-kubernetes-runtime -n oam-system --set image.pullPolicy='Never' --wait \
```
ryanzhang-oss, line 135: IIRC, this is using $(IMAGE), does this work locally (it's pretty handy to debug)?

ryanzhang-oss:
> @ryanzhang-oss I think this is ready now 👍

thx, I just have two questions

Makefile
hasheddan: It is in `image.mk` in the `build/` submodule

Makefile
hasheddan: Yep! Running `make build` will build the image and creates a `values.yaml` file with the image populated :+1:

wonderflow:
@hasheddan  Thanks Daniel !

ryanzhang-oss:
Is the code coverage script working?

PR #92, opened 2020-07-09
[proposal] New workload rollout model support

wonderflow:
In this proposal, I propose that rollout trait should recognize label of workload.


design/workload-rollout-model.md
```diff
@@ -0,0 +1,84 @@
...
+In kubernetes ecosystem, we usually have two categories of rollout model.
+
+1. Workload can rollout by itself: [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) is a typical
+example in this category. When there's a version upgrade, statefulset will directly upgrade pod one by one and allow us to control by the field `partition`.
+2. Workload can't rollout by itself: [ReplicaSet](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/) is a typical example in this model.
+Replica itself don't have rollout feature itself, by we can use it as under layer and build rollout feature by creating two or more replicasets like [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/).
+
+OAM component versioning mechanism has good support for the second category, we regard the workload itself can't rollout, and let
+rollout trait to do it. In this case, if we have a rollout trait which has `revisionEnaled` to be true, oam-runtime will
+create new workload instance with the name to be revisionName. **The revision information passed from oam-runtime to rollout trait from workload name**.
```
ryanzhang-oss, line 22:
```suggestion
create new workload instance with the name to be revisionName. **The workload revision information is then passed from oam-runtime to the rollout trait**.
```

design/workload-rollout-model.md
```diff
@@ -0,0 +1,84 @@
...
+With the revision information, `traffic` trait can route different percent of traffic by specified revisionName.
+
+But in the first category, oam-runtime directly update the workload with the revision information missed. The drawback is we have
```
ryanzhang-oss, line 25: Can you give an example here?

design/workload-rollout-model.md
```diff
@@ -0,0 +1,84 @@
...
+        image: k8s.gcr.io/nginx-slim:0.8
+        ports:
+        - containerPort: 80
+          name: web
+  updateStrategy:
+    partition: 7
+    type: RollingUpdate
+```
+
+For StatefulSet, the old and new version will both exist and with different `oam-revision` label.
```
ryanzhang-oss, line 80: Can you expand the example to have both the old and new version?

resouer:
If I understand it correctly, OAM runtime clearly knows when to add such revision label to workload (i.e. when revision aware trait doesn't exist). 

So in that case, I would suggest we directly assume `PodTemplate` to do the label patch. For workload doesn't have `PodTemplate` I think it's safe assume it can not do self-revision and runtime will ignore this label patch.

wonderflow:
@resouer  We can't assume `PodTemplate` is fully supported, if it's partially supported, this proposal can still work if label exist

wonderflow:
I found pod created by statefulset/cloneset or any workload has revision ability will has a revision label.

```
  apiVersion: v1
  kind: Pod
  metadata:
    labels:
      app: nginx
      controller-revision-hash: frontend-8d6c8c46f
      statefulset.kubernetes.io/pod-name: frontend-2
    name: frontend-2
    namespace: default
```

In this case,  using `controller-revision-hash` label is enough, so oam-runtime won't need to add this label. We can rely on workload itself will have this revision label, and trait can use this label.

wonderflow:
@ryanzhang-oss @resouer  I have updated the proposal that nothing need to change to the OAM spec. Just pointing out the standard way for workload and trait.

design/workload-rollout-model.md
```diff
@@ -0,0 +1,202 @@
...
+For example, traffic trait can specify the label name and detect whether there're different labels exist.
+
+```yaml
+apiVersion: extend.oam.dev/v1alpha2
+kind: Traffic
+spec:
+  revisionLabel: "controller-revision-hash"
+``` 
+ 
+If different labels detected, Traffic Trait should create two service/ingress pointing to different revision labels and
```
ryanzhang-oss, line 197:
```suggestion
If different labels are detected, Traffic Trait could create two service/ingress pointing to different revision labels and
```

wonderflow:
@ryanzhang-oss  Thanks!

Issue #87, opened 2020-07-06
Remove revisions when revisionHistorys over limits


erdun:
There no mechanism that can do cleaning controllerrevisions

ryanzhang-oss:
Thank you for opening this issue, I wonder if you want to create a PR?

erdun:
👌

wonderflow:
fixed by #159

PR #95, opened 2020-07-10
add revision limit

erdun:
Add component's revision limits default 10

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
@@ -30,13 +30,13 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	clientappv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/controller"
 )
 
 const (
@@ -72,16 +72,16 @@ const (
 )
 
 // Setup adds a controller that reconciles ApplicationConfigurations.
-func Setup(mgr ctrl.Manager, l logging.Logger) error {
+func Setup(mgr ctrl.Manager, args controller.Args, l logging.Logger) error {
 	name := "oam/" + strings.ToLower(v1alpha2.ApplicationConfigurationGroupKind)
 
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
 		For(&v1alpha2.ApplicationConfiguration{}).
 		Watches(&source.Kind{Type: &v1alpha2.Component{}}, &ComponentHandler{
-			Client:     mgr.GetClient(),
-			Logger:     l,
-			AppsClient: clientappv1.NewForConfigOrDie(mgr.GetConfig()),
+			Client:               mgr.GetClient(),
+			Logger:               l,
+			DefaultRevisionLimit: args.DefaultRevisionLimit,
 		}).
 		Complete(NewReconciler(mgr,
 			WithLogger(l.WithValues("controller", name)),
@@ -163,11 +163,10 @@ func NewReconciler(m ctrl.Manager, o ...ReconcilerOption) *OAMApplicationReconci
 		client: m.GetClient(),
 		scheme: m.GetScheme(),
 		components: &components{
-			client:     m.GetClient(),
-			appsClient: clientappv1.NewForConfigOrDie(m.GetConfig()),
-			params:     ParameterResolveFn(resolve),
-			workload:   ResourceRenderFn(renderWorkload),
-			trait:      ResourceRenderFn(renderTrait),
+			client:   m.GetClient(),
+			params:   ParameterResolveFn(resolve),
+			workload: ResourceRenderFn(renderWorkload),
+			trait:    ResourceRenderFn(renderTrait),
 		},
 		workloads: &workloads{
 			client:    resource.NewAPIPatchingApplicator(m.GetClient()),
--- a/pkg/controller/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/component.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
@@ -14,7 +15,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	clientappv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -23,11 +23,15 @@ import (
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
 )
 
+// ControllerRevisionComponentLabel indicate which component the revision belong to
+// This label is to filter revision by client api
+const ControllerRevisionComponentLabel = "controller.oam.dev/component"
+
 // ComponentHandler will watch component change and generate Revision automatically.
 type ComponentHandler struct {
-	Client     client.Client
-	AppsClient clientappv1.AppsV1Interface
-	Logger     logging.Logger
+	Client               client.Client
+	Logger               logging.Logger
+	DefaultRevisionLimit int
 }
 
 // Create implements EventHandler
@@ -100,9 +104,16 @@ func (c *ComponentHandler) IsRevisionDiff(mt metav1.Object, curComp *v1alpha2.Co
 	if curComp.Status.LatestRevision == nil {
 		return true, 0
 	}
-	oldRev, err := c.AppsClient.ControllerRevisions(mt.GetNamespace()).Get(context.Background(), curComp.Status.LatestRevision.Name, metav1.GetOptions{})
-	if err != nil {
+
+	// client in controller-runtime will use infoermer cache
+	// use client will be more efficient
+	oldRev := &appsv1.ControllerRevision{}
+	if err := c.Client.Get(context.TODO(), client.ObjectKey{Namespace: mt.GetNamespace(), Name: curComp.Status.LatestRevision.Name}, oldRev); err != nil {
 		c.Logger.Info(fmt.Sprintf("get old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
+		return true, curComp.Status.LatestRevision.Revision
+	}
+	if oldRev.Name == "" {
+		c.Logger.Info(fmt.Sprintf("Not found controllerRevision %s", curComp.Status.LatestRevision.Name), "componentName", mt.GetName())
 		return true, curComp.Status.LatestRevision.Revision
 	}
 	oldComp, err := UnpackRevisionData(oldRev)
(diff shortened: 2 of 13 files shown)


erdun:
Fixed #87

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -36,6 +39,9 @@ func (c *ComponentHandler) Create(evt event.CreateEvent, q workqueue.RateLimitin
 		// No revision created, return
 		return
 	}
+	if err := c.cleanupControllerRevision(evt.Meta, evt.Object); err != nil {
```
ryanzhang-oss, line 42: we only need to do this when we create a revision right? So why not either embed the cleanup logic there or let createControllerRevision return the revision Name to clean? We don't need to list and sort every time

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -47,6 +53,9 @@ func (c *ComponentHandler) Update(evt event.UpdateEvent, q workqueue.RateLimitin
 		// No revision created, return
 		return
 	}
+	if err := c.cleanupControllerRevision(evt.MetaNew, evt.ObjectNew); err != nil {
```
ryanzhang-oss, line 56: same

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -182,6 +191,57 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 	return true
 }
 
+func (c *ComponentHandler) cleanupControllerRevision(mt metav1.Object, obj runtime.Object) error {
```
ryanzhang-oss, line 194: please add UT coverages

apis/core/v1alpha2/core_types.go
```diff
@@ -187,6 +187,11 @@ type ComponentParameter struct {
 
 // A ComponentSpec defines the desired state of a Component.
 type ComponentSpec struct {
+	// RevisionLimit is the maximum number of revisions that will
+	// be maintained. The default value is 10.
+	// +optional
+	RevisionLimit *int32 `json:"revisionLimit,omitempty"`
```
wonderflow, line 193: I'm not sure if we need to make revisionLimit to be component instance level. Maybe system level is enough, that's mean it should be an argument of controller.

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -173,6 +186,17 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 		c.l.Info(fmt.Sprintf("error create controllerRevision %v", err), "componentName", mt.GetName())
 		return false
 	}
+
+	// clean revisions
+	if curComp.Spec.RevisionLimit == nil {
+		curComp.Spec.RevisionLimit = &defaultRevisionLimit
+	}
+	if int64(*curComp.Spec.RevisionLimit) < nextRevision {
+		if err := c.cleanupControllerRevision(mt, curComp); err != nil {
```
wonderflow, line 195: I assume we can create a separate goroutine to do this garbage collection thing, then it won't block the main logic.

wonderflow: cleanupControllerRevision may be time consuming function

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -182,6 +206,63 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+	liveHashes := make(map[string]bool)
+	for _, appConfig := range appConfigs.Items {
+		for _, component := range appConfig.Spec.Components {
+			if component.RevisionName != "" {
+				liveHashes[component.RevisionName] = true
+			}
+		}
+	}
+
+	toKeep := int(*curComp.Spec.RevisionLimit)
```
wonderflow, line 241: I think `toKeep = revisionLimit + len(liveHashes)`

apis/core/v1alpha2/core_types.go
erdun: 👌

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: Goroutine make the process async, that may create conflict， but we can use cache to optimize

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: Livehashes may before the limit or after the limit, if `toKeep = revisionLimit + len(liveHashes)`, the final count may bigger than `toKeep`, by the way if the hash is used, it will be skipped.

apis/core/v1alpha2/core_types.go
erdun: The users are used to RevisionLimit in `Deployment` and `Statefulset`, is it good be an argument of controller?

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: I'll create another PR changing all restful apis to informer machine, and that would modify many places.

pkg/controller/v1alpha2/applicationconfiguration/component.go
wonderflow: RevisionHistory is used be roll back, if `len(liveHashes)` > revisionLimit, we will clean all rest no used revisions, which means we have no revision to roll back.

apis/core/v1alpha2/core_types.go
wonderflow: One of important meanings of OAM is to reduce the complexity of `Deployment` or `Statefulset`

erdun: 👌

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: image
LiveHashes will skipped, don't worry cleaning used revisions.

erdun: 👌

pkg/controller/v1alpha2/applicationconfiguration/component.go
wonderflow: What makes conflict?

erdun: I'm worry about clean revision that used when goroutine delayed and new reconcile triggered.  I'll change it

wonderflow: That won't happen as we have  10 revisions as buffer.

erdun: 👌

wonderflow:
@erdun please help rebase and fix conflict

erdun:
Sorry, I am learning to collaborate in the open source community, please correct me if I'm wrong.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -67,16 +68,17 @@ const (
 )
 
 // Setup adds a controller that reconciles ApplicationConfigurations.
-func Setup(mgr ctrl.Manager, l logging.Logger) error {
+func Setup(mgr ctrl.Manager, args *controller.Args, l logging.Logger) error {
```
wonderflow, line 71: I think it will be more convenient if we directly use struct instead of pointer and won't have risk to get panic.

Because we will use `args.DefaultRevisionLimit` below.

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -173,6 +184,15 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 		c.l.Info(fmt.Sprintf("error create controllerRevision %v", err), "componentName", mt.GetName())
 		return false
 	}
+
+	if int64(c.defaultRevisionLimit) < nextRevision {
+		go func(mt metav1.Object, curComp *v1alpha2.Component) {
```
wonderflow, line 189: If user modify component many times in a short period, I'm afraid we will lose control here.  More general way is setup only one goroutine to do GC, and use a channel to receive event triggered here.

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -182,6 +202,63 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+	revisions, err := c.appsClient.ControllerRevisions(mt.GetNamespace()).List(context.Background(), metav1.ListOptions{
+		LabelSelector: selector.String(),
+	})
+	if err != nil {
+		return err
+	}
+
+	// Get appConfigs and workloads filter controllerRevision used
+	appConfigs := &v1alpha2.ApplicationConfigurationList{}
+	if err := c.client.List(context.Background(), appConfigs); err != nil {
```
wonderflow, line 224: List all appconfig is heavy, what about mark lable on controllerrevison? If AppConfig specify using one revision, mark a lable on controllerrevison object, and this information also can find from status of AppConfig. If someday the revision is removed from AppConfig, we can compare the spec and status to remove the mark from controllerrevison object too.

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: Great!

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
erdun: 👌

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -72,16 +72,16 @@ const (
...
+func Setup(mgr ctrl.Manager, args controller.Args, l logging.Logger) error {
 	name := "oam/" + strings.ToLower(v1alpha2.ApplicationConfigurationGroupKind)
 
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
 		For(&v1alpha2.ApplicationConfiguration{}).
 		Watches(&source.Kind{Type: &v1alpha2.Component{}}, &ComponentHandler{
-			Client:     mgr.GetClient(),
-			Logger:     l,
-			AppsClient: clientappv1.NewForConfigOrDie(mgr.GetConfig()),
```
wonderflow, line 84: It seems AppsClient was removed? This is issue also doing this work. https://github.com/crossplane/oam-kubernetes-runtime/issues/152

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -100,18 +104,25 @@ func (c *ComponentHandler) IsRevisionDiff(mt metav1.Object, curComp *v1alpha2.Co
...
+		return true, curComp.Status.LatestRevision.Revision
+	}
 	oldComp, err := UnpackRevisionData(oldRev)
 	if err != nil {
 		c.Logger.Info(fmt.Sprintf("Unmarshal old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, oldRev.Revision
 	}
 
-	if reflect.DeepEqual(curComp.Spec, oldComp.Spec) {
+	if curComp.Spec.Equal(&oldComp.Spec) {
```
wonderflow, line 125: Why `reflect.DeepEqual` don't work?

erdun: revision.Data read from etcd is  string, for example
`runtime.RawExtension{Object: runtime.Object}` save to etcd
and read from etcd will not decode Object, just be 
`runtime.RawExtension{Raw:[]byte}`
```

wonderflow: that's should be logic contained in refactor client.

erdun: `runtime.RawExtension` has no functions that unMarshal the raw data, and the Object is an interface not a struct, I'm afraid there're no way to undecode the raw data currently.

wonderflow: I suggest we raise another PR to discuss it

erdun: 👌

erdun:
This PR will be closed, I will split into severals.

PR #99, opened 2020-07-14
Add e2e test for dependency engine

My-pleasure:
This e2e test will cover four cases below:
- [x] A trait depends on another trait.
- [x] A component depends on another component.
- [x] A trait depends on a component.
- [x] A component depends on a trait.

The test process for each of these cases consists of two steps:
1. When dependency conditions don't meet requirement, trait or component won't create, and it's unsatisfied dependency will show on `appconfig.status.dependency`.
2. Fill in fieldPath to let conditions meet, trait or component successfully created, and `appconfig.status.dependency` will be nil.

### DIFF
--- a/test/e2e-test/appconfig_dependency_test.go
+++ b/test/e2e-test/appconfig_dependency_test.go
@@ -20,7 +20,7 @@ import (
 	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
 )
 
-var _ = Describe("AppconfigDependency", func() {
+var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 	ctx := context.Background()
 	namespace := "appconfig-dependency-test"
 	var ns corev1.Namespace
--- a/test/e2e-test/appconfig_dependency_test.go
+++ b/test/e2e-test/appconfig_dependency_test.go
@@ -7,7 +7,6 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -15,6 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
 	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
 )
--- a/test/e2e-test/appconfig_dependency_test.go
+++ b/test/e2e-test/appconfig_dependency_test.go
@@ -36,6 +36,8 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 	inName := "data-input"
 	in := tempFoo.DeepCopy()
 	in.SetName(inName)
+	componentOutName := "component-out"
+	componentInName := "component-in"
 	BeforeEach(func() {
 		ns = corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
@@ -89,6 +91,34 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 		logf.Log.Info("Creating trait definition")
 		// For some reason, TraitDefinition is created as a Cluster scope object
 		Expect(k8sClient.Create(ctx, &td)).Should(SatisfyAny(BeNil(), &util.AlreadyExistMatcher{}))
+		// Create a component for data outputs
+		compOut := v1alpha2.Component{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      componentOutName,
+				Namespace: namespace,
+			},
+			Spec: v1alpha2.ComponentSpec{
+				Workload: runtime.RawExtension{
+					Object: out,
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, &compOut)).Should(BeNil())
+		logf.Log.Info("Creating component that outputs data", "Name", compOut.Name, "Namespace", compOut.Namespace)
+		// Create a component for data inputs
+		compIn := v1alpha2.Component{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      componentInName,
+				Namespace: namespace,
+			},
+			Spec: v1alpha2.ComponentSpec{
+				Workload: runtime.RawExtension{
+					Object: in,
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, &compIn)).Should(BeNil())
+		logf.Log.Info("Creating component that inputs data", "Name", compIn.Name, "Namespace", compIn.Namespace)
 	})
 	AfterEach(func() {
 		logf.Log.Info("Clean up resources")
@@ -256,38 +286,6 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 
 	It("component depends on another component", func() {
 		label := map[string]string{"component": "component"}
-		// Create a component
-		componentOutName := "component-comp-out"
-		compOut := v1alpha2.Component{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      componentOutName,
-				Namespace: namespace,
-				Labels:    label,
-			},
-			Spec: v1alpha2.ComponentSpec{
-				Workload: runtime.RawExtension{
-					Object: out,
-				},
-			},
-		}
-		Expect(k8sClient.Create(ctx, &compOut)).Should(BeNil())
-		logf.Log.Info("Creating component that outputs data", "Name", compOut.Name, "Namespace", compOut.Namespace)
-		// Create another component
-		componentInName := "component-comp-in"
-		compIn := v1alpha2.Component{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      componentInName,
-				Namespace: namespace,
-				Labels:    label,
-			},
-			Spec: v1alpha2.ComponentSpec{
-				Workload: runtime.RawExtension{
-					Object: in,
-				},
-			},
-		}
-		Expect(k8sClient.Create(ctx, &compIn)).Should(BeNil())
-		logf.Log.Info("Creating component that inputs data", "Name", compIn.Name, "Namespace", compIn.Namespace)
 		// Create application configuration
 		appConfigName := "appconfig-comp-comp"
 		appConfig := v1alpha2.ApplicationConfiguration{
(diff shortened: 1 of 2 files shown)


ryanzhang-oss:
Can you please fix the build? Are all four cases covered or just the first is?

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+var _ = Describe("AppconfigDependency", func() {
+	ctx := context.Background()
+	namespace := "appconfig-dependency-test"
+	ns := corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		},
+	}
+	BeforeEach(func() {
```
ryanzhang-oss, line 29: You can add some common object here so not every test needs to create OAM resources again

hongchaodeng:
@ryanzhang-oss 
Yeah. I have sync-ed with @My-pleasure yesterday w.r.t. this issue. He is new to this but absolutely working hard to get them fixed..

My-pleasure:
I have covered all four cases and fixed the build. But can't pass the codecov check. /PTAL @ryanzhang-oss @hongchaodeng

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,474 @@
+package controllers_test
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
```
hongchaodeng, line 10: please put this import to below

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,474 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
+)
+
+var _ = Describe("AppconfigDependency", func() {
```
hongchaodeng, line 22:
```suggestion
var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
```

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,474 @@
...
+			Name: namespace,
+		}
+		res := &corev1.Namespace{}
+		Eventually(
+			// gomega has a bug that can't take nil as the actual input, so has to make it a func
+			func() error {
+				return k8sClient.Get(ctx, objectKey, res)
+			},
+			time.Second*30, time.Millisecond*500).Should(&util.NotFoundMatcher{})
+		// recreate it
```
hongchaodeng, line 53:
```suggestion
		// recreate the namespace for testing
```

hongchaodeng:
The codecov is broken... Just ignore it.

ryanzhang-oss:
> I have covered all four cases and fixed the build. But can't pass the codecov check. /PTAL @ryanzhang-oss @hongchaodeng

can you merge with the master and try again? I suspect that the codecov check has a bug

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,475 @@
...
+		logf.Log.Info("Clean up resources")
+		// Delete the WorkloadDefinition and TraitDefinition
+		Expect(k8sClient.Delete(ctx, &wd)).Should(BeNil())
+		Expect(k8sClient.Delete(ctx, &td)).Should(BeNil())
+		// delete the namespace with all its resources
+		Expect(k8sClient.Delete(ctx, &ns, client.PropagationPolicy(metav1.DeletePropagationForeground))).Should(BeNil())
+	})
+
+	// common function for verification
+	verify := func(appConfigName, outName, inName string) {
```
ryanzhang-oss, line 97: This is a good place to use table based testing style

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,475 @@
...
+				Name:      componentName,
+				Namespace: namespace,
+				Labels:    label,
+			},
+			Spec: v1alpha2.ComponentSpec{
+				Workload: runtime.RawExtension{
+					Object: wl,
+				},
+			},
+		}
```
ryanzhang-oss, line 210: dido

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,475 @@
...
+		tdOut := tempFoo.DeepCopy()
+		tdOut.SetName(outName)
+		// Define a trait which accepts data
+		inName := "trait-in"
+		tdIn := tempFoo.DeepCopy()
+		tdIn.SetName(inName)
+		// Define a workload
+		wlName := "wl"
+		wl := tempFoo.DeepCopy()
+		wl.SetName(wlName)
```
ryanzhang-oss, line 196: you don't need to define this in every test, put them in one test, or put those in the parent scope so you can save all the repetitions.

hongchaodeng:
> can you merge with the master and try again? I suspect that the codecov check has a bug that it calculated the coverage purely on the branch vs the master instead of merging the branch first.

We shouldn't require contributors to merge master every time the master gets updated. If it's a bug in codecov, then we should just stop using codecov in CI.

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,475 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+
+	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
```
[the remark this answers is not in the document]
hongchaodeng: AFAICT, this is already gofmt-ed. The CI also has checks for that. What are you expecting to change here?

test/e2e-test/appconfig_dependency_test.go
hongchaodeng: Ginkgo promotes BDD style and this is the correct way to do it.
AFAIK, TBT should not be used here otherwise the two styles would just conflict.

test/e2e-test/appconfig_dependency_test.go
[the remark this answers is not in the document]
ryanzhang-oss: There should not be 5 import groups

test/e2e-test/appconfig_dependency_test.go
ryanzhang-oss: I don't think Ginkgo conflicts with TBT but I am ok with either style, just not to copy-paste the same setup code again and again.

ryanzhang-oss:
> > can you merge with the master and try again? I suspect that the codecov check has a bug that it calculated the coverage purely on the branch vs the master instead of merging the branch first.
> 
> We shouldn't require contributors to merge master every time the master gets updated. If it's a bug in codecov, then we should just stop using codecov in CI.

Oh,  it's just my theory, we should try to solve the codecov bug or replace it with something else. I don't think we can get away with code coverage gate in CI though.

hongchaodeng:
> Oh, it's just my theory, we should try to solve the codecov bug or replace it with something else.

I agree. But this is not something this PR should try to resolve. Can you create an issue to resolve it later?

>  I don't think we can get away with code coverage gate in CI though.

It's already not required in CI gate.

test/e2e-test/appconfig_dependency_test.go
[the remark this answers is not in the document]
hongchaodeng: > There should not be 5 import groups

What's the reason?

test/e2e-test/appconfig_dependency_test.go
hongchaodeng: >  just not to copy-paste the same setup code again and again.

Totally agree.

test/e2e-test/appconfig_dependency_test.go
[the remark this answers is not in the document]
hongchaodeng: I have tried hard with the code checks before so I can guess this is what @My-pleasure tried:

- go built-in should be together in one block
- ginkgo are imported as `. xxx` so they are in one block
- external deps are in one block
- `crossplane/crossplane-runtime` are in one block otherwise the CI will fail
- `crossplane/oam-kubernetes-runtime/` in one block otherwise the CI will fail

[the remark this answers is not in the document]
hongchaodeng: This seems valid to me as dealing with the import rules from the CI would lead to this result..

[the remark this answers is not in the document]
ryanzhang-oss: Have you tried to use `goimports`, it's the same idea as `gofmt` so we don't need to talk about imports in reviews.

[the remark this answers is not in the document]
hongchaodeng: I just tried to checkout the code and run goimports. It didn't print out any diff so it concludes that this is correctly formatted.

test/e2e-test/appconfig_dependency_test.go
My-pleasure: I have put the same setup codes to the parent scope. /PTAL

test/e2e-test/appconfig_dependency_test.go
My-pleasure: I'm sorry I don't get what you mean

hongchaodeng: Explained offline to @My-pleasure

My-pleasure: In the "trait depend on trait" case, I think we need an extra component. And in other cases, I have already put component definitions to the BeforeEach block.

test/e2e-test/appconfig_dependency_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(k8sClient.Update(ctx, outFoo)).Should(Succeed())
+		// Verification after satisfying dependency
+		By("Checking that resource which accepts data is created now")
+		logf.Log.Info("Checking on resource that inputs data", "Key", inFooKey)
+		Eventually(
+			func() error {
+				return k8sClient.Get(ctx, inFooKey, inFoo)
+			},
+			time.Second*15, time.Millisecond*500).Should(BeNil())
+		By("Verify the appconfig's dependency is unsatisfied")
```
ryanzhang-oss, line 208: satisfied?

My-pleasure: I have fixed the logging error now. @ryanzhang-oss

PR #102, opened 2020-07-15
Move Dockerfile to images dir for compatability with build submodule

hasheddan:
I believe this will fix the issue of the image and helm chart not being published.


hasheddan:
@ryanzhang-oss could I get a review here :+1:

PR #106, opened 2020-07-17
Enable workloadRefPath field in ScopeDefinition

### LINKED ISSUE #69: Do not assume that a scope has a workloadRef field
AppConfig now hardcode the workloadRef field on a scope CR. We need to use the same mechanism as trait to know the workloads it is associated with.

captainroy-hy:
fixes #69 
add new field in apis
update crd of ScopeDefinition
update sample HealthScope's ScopeDefinition
add unit test cases


apis/core/v1alpha2/core_types.go
```diff
@@ -125,6 +125,9 @@ type ScopeDefinitionSpec struct {
 	// Reference to the CustomResourceDefinition that defines this scope kind.
 	Reference DefinitionReference `json:"definitionRef"`
 
+	// WorkloadRefPath indicates if/where a scope accepts a workloadRef object
+	WorkloadRefPath string `json:"workloadRefPath,omitempty"`
```
hongchaodeng, line 129: For scope, it should be an array of references:
https://github.com/crossplane/oam-kubernetes-runtime/blob/df8e1b1c6beeb38a9a0d3ad16057d157eab9984d/apis/core/v1alpha2/core_scope_types.go#L34

captainroy-hy: That's right. I will fix it.

design/one-pager-scope-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,141 @@
+# Scopes and workloads interaction mechanism in OAM
+
+* Owner: Ryan Zhang (@ryanzhang-oss)
+* Status: Implemented
```
wonderflow, line 4: I think it's copy-paste mistake here.

design/one-pager-scope-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,141 @@
...
+            apiVersion: core.oam.dev/v1alpha2
+            kind: HealthScope
+            name: example-health-scope
+```
+
+The problem is two folds
+1. A scope controller needs a way to find the workloads that it's going to control. 
+   - In the example, HealthScope should know that it is supposed to check the health condition of `my-example-workload`. 
+However, we want to keep the applicationConfiguration controller agnostic to the schema of any `scope` it interacts with or `workload` it generates to make it extensible. 
+Thus, the applicationConfiguration controller needs to get the HealthScope CR (because scope is a kind of global resource) and give it a reference of the `my-example-workload` workload **without knowing HealthScope's specific schema**.
```
wonderflow, line 69: We have already implemented `workloadRefs` in scope which solves this problem, I think we should narrow the problem to be `current implementation is hard code` issue.

design/one-pager-scope-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,141 @@
+# Scopes and workloads interaction mechanism in OAM
```
ryanzhang-oss, line 1: This is very similar to the trait-workload interaction, how about just modify that proposal instead?

pkg/oam/util/helper.go
```diff
@@ -79,6 +79,21 @@ func LocateParentAppConfig(ctx context.Context, client client.Client, oamObject
 	return nil, errors.Errorf(ErrLocateAppConfig)
 }
 
+// FetchScopeDefinition fetch corresponding scopeDefinition given a scope
+func FetchScopeDefinition(ctx context.Context, r client.Reader,
```
ryanzhang-oss, line 83: no test?

captainroy-hy: sorry for overlooking. I will add a test for it.

design/one-pager-scope-workload-interaction-mechanism.md
captainroy-hy: maybe I can follow Ryan's idea 
- just modify the existing proposal instead of a new one, 

and follow @wonderflow advice to 
- write a separate section from Trait's in the proposal.
- narrow the problem statement to be `current implementation is hard code` issue

design/one-pager-scope-workload-interaction-mechanism.md
captainroy-hy: not exactly... I think Ryan came up with the proposal and I just follow to update codes.

design/one-pager-scope-workload-interaction-mechanism.md
captainroy-hy: I have merged this proposal into existing one. Once you check it as okay, I will remove the separate one and rename the existing one, e.g. "one-pager-trait-scope-workload-interaction-mechanism.md"

PR #116, opened 2020-07-21
enable extension ability of definition for OAM platform builder

wonderflow:
Add extension field into WorkloadDefinition, TraitDefinition and ScopeDefinition.


apis/core/v1alpha2/core_types.go
```diff
@@ -49,6 +50,10 @@ type WorkloadDefinitionSpec struct {
 
 	// ChildResourceKinds are the list of GVK of the child resources this workload generates
 	ChildResourceKinds []ChildResourceKind `json:"childResourceKinds,omitempty"`
+
+	// Extension is used for extension needs by OAM platform builders
+	// +optional
+	Extension *unstructured.Unstructured `json:"extension,omitempty"`
```
hongchaodeng, line 56: I think one Unstructured as extension might not be enough. We should design it for more future extension needs.

How about change it to:
```
Extensions map[string]*Unstructured
```

So we could have:

```
extensions:
  template:
    ...
```

wonderflow: Current struct can also represent. The underlying go structure is map[string]interface{} which means can represent any object.

```
extensions:
  template:
    ...
```

resouer:
@wonderflow Please also send a PR to the spec repo to reflect this change (as part of v1alpha2 spec).

PR #119, opened 2020-07-22
init admission webhook framework for validation

### LINKED ISSUE #73: We need an admission controller
Currently, an applictionConfig controller does not do many validations. It is less user friendly for the controller to fail to reconcile than we reject any invalid OAM applications. An admission controller is exactly for that purpose.

### LINKED ISSUE #81: replace gcr.io image to docker hub
gcr.io is blocked in some area, so we'd better use images from docker hub instead.

https://github.com/crossplane/oam-kubernetes-runtime/blob/master/charts/oam-core-runtime/templates/oam-controller.yaml#L114

wonderflow:

bootstrap admission webhook framework for validation  fixes #73

- [x] update helm chart: support useWebhook as values, clean no used configs
- [x] add README to teach how to use webhook
- [x] add certs generate script instead of using certmanager, as certmanager costs too much resources
- [x] replace grc.io images : fixes #81
- [x] add validation for revisionName/componentName are mutually exclusive
- [x] test coverage


cmd/oam-kubernetes-runtime/main.go
```diff
@@ -26,6 +29,13 @@ func init() {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
+	var certDir string
+	var webhookPort int
+	var useWebhook bool
+
+	flag.BoolVar(&useWebhook, "use-webhook", false, "Enable Admission Webhook")
```
hongchaodeng, line 36: this should be `enable-webhook`?

ryanzhang-oss:
why not use cert-manager which seems to be the defacto cert generator

wonderflow:
@ryanzhang-oss  certmanager costs too much resources

cmd/oam-kubernetes-runtime/main.go
wonderflow: OAM chart use this word, so I followed it. I think that won't cause confusion

README.md
```diff
@@ -33,6 +33,39 @@ helm repo add crossplane-master https://charts.crossplane.io/master/
 helm install oam --namespace oam-system crossplane-master/oam-kubernetes-runtime --devel
 ```
 
+## Install OAM Runtime with Webhook
+
+1. Admission Webhook need you to prepare certificates and ca for production use.
```
ryanzhang-oss, line 38: sounds like someone has to clone the repo only if they want to use the webhook?

README.md
```diff
@@ -33,6 +33,39 @@ helm repo add crossplane-master https://charts.crossplane.io/master/
...
+**For none-production use**, you cloud generate it by the shell script provided in repo.
```
ryanzhang-oss, line 39:
```suggestion
**For none-production use**, you could generate them by running the shell script provided in repo.
```

charts/oam-kubernetes-runtime/templates/webhook.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.useWebhook -}}
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
```
ryanzhang-oss, line 3: why validating only?

charts/oam-kubernetes-runtime/templates/webhook.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - apiGroups:   ["core.oam.dev"]
+        apiVersions: ["v1alpha2"]
+        operations:  ["CREATE", "UPDATE","DELETE"]
+        resources:   ["applicationconfigurations"]
+        scope:       "Namespaced"
+    clientConfig:
+      service:
+        namespace: {{.Release.Namespace}}
+        name: {{ template "oam-kubernetes-runtime.name" . }}-webhook
+        path: /validating-applicationconfigurations
```
ryanzhang-oss, line 21: is there any way to make sure this path is the same as the path in the below function

func Register(server *webhook.Server) {
	server.Register("/validating-applicationconfigurations", &webhook.Admission{Handler: &ValidatingHandler{}})
}

wonderflow: No other way, we can configure it right. It's configurable here

charts/oam-kubernetes-runtime/templates/webhook.yaml
wonderflow: No mutating needs

README.md
wonderflow: yes, only if they want to use our shell to generate

PR #121, opened 2020-07-22
Generate v1 CRD instead of v1beta1

### LINKED ISSUE #118: Generate v1 CRD instead of v1beta1
# Motivation
v1beta1 CRD has been deprecated in k8s 1.16+ . That means now v1beta1 CRD isn't supported anymore.

However, currently all of our CRDs are generated as v1beta1:
https://github.com/crossplane/oam-kubernetes-runtime/blob/b678ec6a4302b458349651211247847fc829b3d0/charts/oam-kubernetes-runtime/crds/core.oam.dev_applicationconfigurations.yaml#L3

We should switch to generate v1 CRD instead.

# How

The CRDs are generated in this code:
https://github.com/crossplane/oam-kubernetes-runtime/blob/335f6fea2658cec1bc776f57a47c06be98ce7f1a/apis/generate.go#L26

We need to add `crd:crdVersions=v1` to it.

Additionally, we should remove `crd:trivialVersions` tag as we are adding conversion webhook. 
More info are available in the [code](https://github.com/kubernetes-sigs/controller-tools/blob/b45abdba0dbb53e0c4dc90e1082d3802ecfb8be4/pkg/crd/gen.go#L38-L45)

My-pleasure:
- Modify `crd:trivialVersions` tag to `crd:crdVersions=v1` in generate.go file.
- Regenerate the CRDs with version v1.

fixes #118

### DIFF
--- a/apis/generate.go
+++ b/apis/generate.go
@@ -23,7 +23,7 @@ limitations under the License.
 //go:generate rm -rf ../charts/oam-kubernetes-runtime/crds
 
 // Generate deepcopy methodsets and CRD manifests
-//go:generate go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=../hack/boilerplate.go.txt paths=./... crd:trivialVersions=true output:artifacts:config=../charts/oam-kubernetes-runtime/crds
+//go:generate go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=../hack/boilerplate.go.txt paths=./... crd:crdVersions=v1 output:artifacts:config=../charts/oam-kubernetes-runtime/crds
 
 package apis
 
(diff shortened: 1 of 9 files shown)


hasheddan:
I don't think this should be done yet. 1.16 is not yet out of support until 1.19 is released and many hosted Kubernetes providers still offer 1.15 and earlier versions. At the least, both should be generated and users should be able to decide whether they want v1beta1 or v1 CRDs installed using a flag on the helm chart.

hongchaodeng:
Hi @hasheddan 

> 1.16 is not yet out of support

Whether or not 1.16 i out of support doesn't matter. It is 1.15 that is out of support. Right?
If 1.15 is out of support, that means v1beta1 CRD is out of support.

> many hosted Kubernetes providers still offer 1.15 and earlier versions.

Why do we care about that? If some providers offer out of support versions it is their jobs to maintain that. The open source community will keep track of upstream (i.e. k8s) only.

> At the least, both should be generated and users should be able to decide whether they want v1beta1 or v1 CRDs installed using a flag on the helm chart.

I'm fine with this option. But it should not be default.
For upstream, we should not support any deprecated versions. Otherwise it just adds more to our burden.

PR #105, opened 2020-07-17
add pre and post hooks for customized reconcile logic

ryanzhang-oss:
- add extra status fields for customized reconcile
- add pre and post hooks for customized reconcile logic

PR #127, opened 2020-07-23
Fix nil panic in WithPrehook and WithPosthook

Diddaa:


Diddaa:
relate to pr #105

ryanzhang-oss:
it's fixed in dev branch

PR #134, opened 2020-07-23
implement Trait patch back before workload emitted

ifree613:
Ref to proposal https://github.com/crossplane/oam-kubernetes-runtime/blob/master/design/trait-separated-from-underlying-workload-should-patch-back.md

Signed-off-by: chengji.lsb <chengji.lsb@alibaba-inc.com>


pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -173,26 +194,81 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
...
+	nn := types.NamespacedName{Namespace: w.GetNamespace(), Name: patchName}
+	cm := &corev1.ConfigMap{}
+	var patch string
+	if err := r.client.Get(ctx, nn, cm); err == nil {
+		if patch = cm.Data[patchKey]; patch == "" {
+			return errors.Wrapf(err, errGetConfigMapValueForName, nn.Name)
+		}
+	} else {
+		return err
+	}
```
wonderflow, line 222:
```suggestion
	if err := r.client.Get(ctx, nn, cm); err != nil {
		return err
	}
        if patch = cm.Data[patchKey]; patch == "" {
	        return errors.Wrapf(err, errGetConfigMapValueForName, nn.Name)
        }
```

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -173,26 +194,81 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
...
+
+	workloadFullJSON, err := jsonpatch.MergeMergePatches(workloadJSON, []byte(patch))
+	if err != nil {
+		return err
+	}
+
+	var mergePatchObject map[string]interface{}
+	if err = json.Unmarshal(workloadFullJSON, &mergePatchObject); err == nil {
```
wonderflow, line 235: We are used to judge if err != nil  first

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -323,6 +399,9 @@ func renderWorkload(data []byte, p ...Parameter) (*unstructured.Unstructured, er
 			// TODO(negz): Infer parameter type from workload OpenAPI schema.
 			switch param.Value.Type {
 			case intstr.String:
+				if path == fmt.Sprintf("%s.%s", StatusKey, patchConfigKey) {
```
wonderflow, line 402: Add a flag for internal use and check here would be better

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -56,6 +58,16 @@ const (
...
+	errGetConfigMapValueForName  = "can not get configMap data for configMapName %q"
+	errNoSuchField               = "status: no such field"
+)
+
+const (
+	//StatusKey stands for the status field of the spec
+	StatusKey      = "status"
+	patchConfigKey = "patchConfig"
+	patchKey       = "patch"
+	patchNameFmt   = "%s-cm"
```
wonderflow, line 70: we shouldn't assume patchName has this suffix

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -172,26 +193,87 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 	return &Workload{ComponentName: acc.ComponentName, ComponentRevisionName: componentRevisionName, Workload: w, Traits: traits, Scopes: scopes}, nil
 }
 
-func (r *components) renderTrait(ctx context.Context, ct v1alpha2.ComponentTrait, ac *v1alpha2.ApplicationConfiguration,
+func (r *components) renderComponentFromPatchTrait(ctx context.Context, w *unstructured.Unstructured, acc *v1alpha2.ApplicationConfigurationComponent, t *unstructured.Unstructured) error {
```
ryanzhang-oss, line 196: Does this one have a unit test? I would also suggest to break this function into multiple pieces (the patch itself can be a function)

ifree613: OK, I'll add a unit test

hongchaodeng:
Any progress on this PR?
If @ifree613 doesn't have time, we can help continue to finish it.
cc @wonderflow

wonderflow:
pending on this issue https://github.com/crossplane/oam-kubernetes-runtime/issues/157

ryanzhang-oss:
Just FYI, the strategicMergepatch is in a library at 	"k8s.io/apimachinery/pkg/util/strategicpatch"

antiClocke:
//oam-k8s-runtime will periodically check the status of patch trait. If the phase of status become Ready, the oam-k8s-runtime will gather the patch data from the ConfigMap. After all patch traits gathered, oam-k8s-runtime will patch all the information into workload

antiClocke:
phase: Ready

antiClocke:
//All patch traits MUST be created or updated before workload emit, and oam-k8s-runtime will be pending until all patch traits are ready.
i am a fish,a questioin
where is the pending logic, i d not fount

wonderflow:
@antiClocke Let's talk deep in wechat or Dingtalk, my ID is  wonderflow

antiClocke:
> @antiClocke Let's talk deep in wechat or Dingtalk, my ID is wonderflow

enen , wechat

PR #135, opened 2020-07-23
Fix unstable e2e test of dependency engine

My-pleasure:
Because the instability of e2e test is mainly in these two places after the dependency conditions are satisfied:
1. the update of appconfig
2. the start of resource which accepts data.

So I expand Eventually's timeout and polling interval in these two places.

(By the way, I update CRD version to v1 instead of v1beta1.)

### DIFF
--- a/test/e2e-test/appconfig_dependency_test.go
+++ b/test/e2e-test/appconfig_dependency_test.go
@@ -204,7 +204,7 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 			func() error {
 				return k8sClient.Get(ctx, inFooKey, inFoo)
 			},
-			time.Second*15, time.Millisecond*500).Should(BeNil())
+			time.Second*60, time.Second*5).Should(BeNil())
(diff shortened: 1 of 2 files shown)


test/e2e-test/appconfig_dependency_test.go
```diff
@@ -204,7 +204,7 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 			func() error {
 				return k8sClient.Get(ctx, inFooKey, inFoo)
 			},
-			time.Second*15, time.Millisecond*500).Should(BeNil())
+			time.Second*60, time.Second*5).Should(BeNil())
```
wonderflow, line 207: polling interval set to 1s would be better?

My-pleasure: ok, and then the timeout value should be bigger? Because last time, it failed when timeout is 60s, polling interval is 5s.

hongchaodeng: 5s should be fine.

Issue #138, opened 2020-07-24
Update readme docs in examples


wonderflow:
Now we have released helm chart for oam-k8s-runtime, and `addon-oam-kubernetes-local` already merged in `oam-k8s-runtime`. We should update docs to align with[ the installation](https://github.com/crossplane/oam-kubernetes-runtime#install-oam-runtime).

I have found at least these places need to update:

* https://github.com/crossplane/oam-kubernetes-runtime/tree/master/examples/dependency#prerequisite
* https://github.com/crossplane/oam-kubernetes-runtime/blob/master/examples/component-versioning/versioning-mechanism.md#prerequisite
* https://github.com/crossplane/oam-kubernetes-runtime/blob/master/examples/component-versioning/component-mutable.md#prerequisite

PR #142, opened 2020-07-26
fix examples with the new code

leopoldxx:
the examples has been broken because of the evolvement of the code.  fix the examples with the new code.

fix: https://github.com/crossplane/oam-kubernetes-runtime/issues/138

### DIFF
--- a/examples/component-versioning/component-mutable.md
+++ b/examples/component-versioning/component-mutable.md
@@ -2,9 +2,7 @@
 
 ## Prerequisite
 
-1. Make sure [`addon-oam-kubernetes-local`](https://github.com/crossplane/addon-oam-kubernetes-local) was installed.
-2. Before component versioning mechanism released in Crossplane, you can run `go run examples/containerized-workload/main.go`
-   instead of using crossplane. After this feature was released in crossplane, install crossplane will work.
+  Make sure [`OAM runtime`](../../README.md#install-oam-runtime) was installed and started.
 
 ## ApplicationConfiguration always using the latest component
 
@@ -14,7 +12,7 @@ Step 1. Create OAM component
 $ kubectl apply -f examples/containerized-workload/sample_workload_definition.yaml
 $ kubectl apply -f examples/containerized-workload/sample_trait_definition.yaml
 $ kubectl apply -f examples/containerized-workload/sample_scope_definition.yaml
-$ kubectl apply -f examples/containerized-workload/sample_component.yaml
+$ kubectl apply -f examples/component-versioning/sample_component.yaml
 component.core.oam.dev/example-component created
 ``` 
 
@@ -32,6 +30,8 @@ Snapshot of current component stored in this ControllerRevision
 $ kubectl get controllerrevisions.apps example-component-brk71rbipt3d60vfo4t0 -o yaml
 apiVersion: apps/v1
 data:
+  apiVersion: core.oam.dev/v1alpha2
+  kind: Component
   metadata:
     name: example-component
     namespace: default
@@ -51,7 +51,9 @@ data:
           - containerPort: 80
             name: wordpress
   status:
-    latestRevision: ""
+    latestRevision:
+      name: example-component-bsemlvvoceaedeu80kfg
+      revision: 1
 kind: ControllerRevision
 metadata:
   name: example-component-brk71rbipt3d60vfo4t0
@@ -68,7 +70,7 @@ revision: 1
 Step 2. Create AppConfig
 
 ```shell script
-$ kubectl apply -f examples/containerized-workload/sample_application_config.yaml
+$ kubectl apply -f examples/component-versioning/sample_application_config.yaml
 applicationconfiguration.core.oam.dev/example-appconfig created
 ```
 
@@ -188,7 +190,7 @@ component.core.oam.dev "example-component" deleted
 Step 1. The first step is the same. Create OAM component, and check the ControllerRevision.
 
 ```shell script
-$ kubectl apply -f examples/containerized-workload/sample_component.yaml
+$ kubectl apply -f examples/component-versioning/sample_component.yaml
 component.core.oam.dev/example-component created
 ``` 
 
@@ -219,10 +221,10 @@ spec:
               replicaCount: 3
 ```
 
-Assume we name it as `component-mutable-app.yaml` and apply this AppConfig.
+Assume we name it as `examples/component-versioning/component-mutable-app.yaml` and apply this AppConfig.
 
 ```shell script
-$ kubectl apply -f component-mutable-app.yaml
+$ kubectl apply -f examples/component-versioning/component-mutable-app.yaml
 applicationconfiguration.core.oam.dev/example-appconfig created
 ```
 
@@ -270,6 +272,7 @@ component.core.oam.dev/example-component edited
 The controllerRevision was created.
 
 ```shell script
+$ kubectl get controllerrevisions.apps
 NAME                                     CONTROLLER                                 REVISION   AGE
 example-component-brk71b3ipt3d60vfo4sg   component.core.oam.dev/example-component   1          29m
 example-component-brke6rbipt3d60vfo4ug   component.core.oam.dev/example-component   2          73s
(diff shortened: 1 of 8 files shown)


examples/component-versioning/component-mutable.md
```diff
@@ -2,9 +2,11 @@
...
 ## Prerequisite
 
-1. Make sure [`addon-oam-kubernetes-local`](https://github.com/crossplane/addon-oam-kubernetes-local) was installed.
-2. Before component versioning mechanism released in Crossplane, you can run `go run examples/containerized-workload/main.go`
-   instead of using crossplane. After this feature was released in crossplane, install crossplane will work.
+   Start OAM runtime:
+
+   ```shell script
+    go run cmd/oam-kubernetes-runtime/main.go
+   ```
```
wonderflow, line 9: Ref to in [the installation step](https://github.com/crossplane/oam-kubernetes-runtime#install-oam-runtime) insteand of `go run`.

examples/component-versioning/component-mutable.md
```diff
@@ -37,6 +39,13 @@ data:
     namespace: default
     ...
   spec:
+    parameters:
+    - fieldPaths:
+      - metadata.name
+      name: instance-name
+    - fieldPaths:
+      - spec.containers[0].image
+      name: image
```
wonderflow, line 48: I don't think these parameters are needed. Instead, using component revisioning discourage users using parameters.

examples/component-versioning/component-mutable.md
```diff
@@ -95,7 +106,7 @@ spec:
   - env:
     - name: TEST_ENV
       value: test
-    image: wordpress:4.6.1-apache
+    image: wordpress:php7.2
```
wonderflow, line 109: Why make this tag change?

examples/dependency/README.md
```diff
@@ -10,7 +10,7 @@ kubectl apply -f examples/dependency/definition.yaml
 Start OAM runtime:
 
 ```shell script
-go run ./cmd/oam-runtime/main.go
+go run cmd/oam-kubernetes-runtime/main.go
```
wonderflow, line 13: Using [the installation step](https://github.com/crossplane/oam-kubernetes-runtime#install-oam-runtime)

examples/dependency/demo-with-conditions.yaml
```diff
@@ -43,14 +43,4 @@ spec:
...
-      traits:
-        - trait:
-            apiVersion: serving.knative.dev/v1
-            kind: Route
-            metadata:
-              name: trait-test
-            spec:
-              traffic:
-                - configurationName: webapp
-                  percent: 100
```
wonderflow, line 56: why deleting?

examples/dependency/README.md
```diff
@@ -155,6 +155,24 @@ After a while, assuming the controller of this CRD will reconcile and make the `
...
+    kind: Foo
+    metadata:
+      name: source
+    spec:
+      key: test-value
+-   status:
+-      state: pending
++ #   status:
++ #      state: pending
+
```
wonderflow, line 175: why edit component ?

examples/component-versioning/component-mutable.md
leopoldxx: I run the examples and get that result.  I think it comes from https://github.com/crossplane/oam-kubernetes-runtime/blob/55cd9dbe93cfeba4fe03dfd8c81c34a4f367d2bc/examples/containerized-workload/sample_component.yaml#L19

examples/component-versioning/component-mutable.md
leopoldxx: the original `sample_application_config` example modified the image parameter, this is the result of the parameter render: 
https://github.com/crossplane/oam-kubernetes-runtime/blob/55cd9dbe93cfeba4fe03dfd8c81c34a4f367d2bc/examples/containerized-workload/sample_application_config.yaml#L8-L12

examples/dependency/demo-with-conditions.yaml
leopoldxx: It is lack of `serving.knative` support here,  and the example won't work if the user experiments the example. 

Or it is needed to add some knative support in prerequisite section, but i don't know if the `Route` trait can work correctly on this workload. 

So should i restore this section?

examples/dependency/README.md
leopoldxx: this component yaml set `.status.state` to `pending`.

if we only edit `source` workload's `.status.state` and set it to `running`,  then it will be reconciled to `pending` state according to component's value soon.

examples/component-versioning/component-mutable.md
wonderflow: I see, we should copy the sample_component yaml from containerized-workload and make sure the new one without the parameters section. The the image tag below won't changed.

examples/dependency/README.md
wonderflow: I can't very understand here. According to the context, the `status.state` is [already pending](https://github.com/crossplane/oam-kubernetes-runtime/pull/142/files#diff-c4f5f305c3bca0c5fea1e4f20a8d5ad2R137-R138) , and in your example, your are removing the field?

examples/dependency/demo-with-conditions.yaml
wonderflow: I see, this section should be deleted.

examples/component-versioning/component-mutable.md
leopoldxx: `containerized-workload/sample_component.yaml` has this `parameters` section,  so `copy` won't work.  should we delete that code  from `containerized-workload/sample_component.yaml`  ?https://github.com/crossplane/oam-kubernetes-runtime/blob/55cd9dbe93cfeba4fe03dfd8c81c34a4f367d2bc/examples/containerized-workload/sample_component.yaml#L1-L26

examples/dependency/README.md
leopoldxx: `.status.state` is set to `pending` in  Component `source`. 
https://github.com/crossplane/oam-kubernetes-runtime/blob/55cd9dbe93cfeba4fe03dfd8c81c34a4f367d2bc/examples/dependency/demo-with-conditions.yaml#L7-L15

Even we modify `.status.state` of workload Foo `source` and set the value to `running`, it will be re-rendered to `pending` according to  Component `source`'s definition(by application controller). 

i will re-experiment this use case  to check it.

examples/component-versioning/component-mutable.md
wonderflow: We can copy that, and remove the parameters section in the new example.

examples/component-versioning/component-mutable.md
```diff
@@ -95,7 +102,7 @@ spec:
...
```
wonderflow, line 105: if we don't have parameter, then the change won't happen.

leopoldxx: i see, when finish the modification above (https://github.com/crossplane/oam-kubernetes-runtime/pull/142#discussion_r461350788) , i will restore this part.

examples/component-versioning/component-mutable.md
leopoldxx: fixed.

examples/component-versioning/component-mutable.md
leopoldxx: fixed

examples/dependency/README.md
leopoldxx: I execute this case again, yes, editing Foo `source`'s `state` is enough, and the Foo `sink` object will be create after the modification. but Foo `source`'s `state`  will also be reconciled back to `pending` because of Component `source`'s `state` value. the key is which come first.  

i will remove the this  component editing part.

leopoldxx:
@wonderflow

I don't understand why the e2e test failed?  I review this test code and think the modification of the pr should not affect the test. 

and how to re-run the test?  

thanks for review.

wonderflow:
@leopoldxx That's not your fault, the e2e test still has some problem, let me create an issue.

wonderflow:
@leopoldxx  It seems e2e was fixed, can you help rebase this PR?

leopoldxx:
> @leopoldxx It seems e2e was fixed, can you help rebase this PR?

@wonderflow done.

PR #148, opened 2020-07-28
Add webhook validation for component versioning

captainroy-hy:
- Add validation in webhook for component versioning mechanism
- Add unit tests
- Init webhook README.md

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/component.go
@@ -21,7 +21,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
-	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
 )
 
 // ComponentHandler will watch component change and generate Revision automatically.
@@ -106,7 +105,7 @@ func (c *ComponentHandler) IsRevisionDiff(mt metav1.Object, curComp *v1alpha2.Co
 		c.Logger.Info(fmt.Sprintf("get old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, curComp.Status.LatestRevision.Revision
 	}
-	oldComp, err := util.UnpackRevisionData(oldRev)
+	oldComp, err := UnpackRevisionData(oldRev)
 	if err != nil {
 		c.Logger.Info(fmt.Sprintf("Unmarshal old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, oldRev.Revision
--- a/pkg/controller/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/component.go
@@ -21,6 +21,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
 )
 
 // ComponentHandler will watch component change and generate Revision automatically.
@@ -105,7 +106,7 @@ func (c *ComponentHandler) IsRevisionDiff(mt metav1.Object, curComp *v1alpha2.Co
 		c.Logger.Info(fmt.Sprintf("get old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, curComp.Status.LatestRevision.Revision
 	}
-	oldComp, err := UnpackRevisionData(oldRev)
+	oldComp, err := util.UnpackRevisionData(oldRev)
 	if err != nil {
 		c.Logger.Info(fmt.Sprintf("Unmarshal old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, oldRev.Revision
--- a/pkg/oam/util/helper_test.go
+++ b/pkg/oam/util/helper_test.go
@@ -12,13 +12,13 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/test"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	gomegaTypes "github.com/onsi/gomega/types"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	fakeClient "k8s.io/client-go/kubernetes/fake"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -892,13 +892,16 @@ var _ = Describe("Test get component helper utils", func() {
 			acc                v1alpha2.ApplicationConfigurationComponent
 			expectComponent    *v1alpha2.Component
 			expectRevisionName string
-			expectError        error
+			expectErrorMatcher gomegaTypes.GomegaMatcher
 		}
 
 		namespace := "ns"
 		componentName := "newcomponent"
+		invalidComponentName := "invalidComponent"
 		revisionName := "newcomponent-aa1111"
 		revisionName2 := "newcomponent-bb1111"
+		unpackErrorRevisionName := "unpackErrorRevision"
+		errComponentNotFound := errors.New("component not found")
 
 		componnet1 := v1alpha2.Component{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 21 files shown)


pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -65,6 +75,40 @@ func checkRevisionName(appConfig *v1alpha2.ApplicationConfiguration) (bool, stri
 	return true, ""
 }
 
+// checkWorkloadNameForVersioning check whether versioning-enabled component workload name is empty
+func checkWorkloadNameForVersioning(ctx context.Context, client client.Reader, appConfig *v1alpha2.ApplicationConfiguration) (bool, string) {
```
ryanzhang-oss, line 79: why return a string instead of error

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -0,0 +1,141 @@
...
+	nk := types.NamespacedName{Namespace: namespace, Name: acc.ComponentName}
+	if err := client.Get(ctx, nk, c); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+type parameter acCtrl.Parameter
+
+func resolveParams(cp []v1alpha2.ComponentParameter, cpv []v1alpha2.ComponentParameterValue) ([]parameter, error) {
```
ryanzhang-oss, line 79: are you sure that this is no similar function in the appConfig controller?

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -0,0 +1,141 @@
...
+
+	params := make([]parameter, 0, len(set))
+	for _, p := range set {
+		params = append(params, *p)
+	}
+
+	return params, nil
+}
+
+func renderWorkload(data []byte, p ...parameter) (*unstructured.Unstructured, error) {
```
ryanzhang-oss, line 117: why copy code here?

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -0,0 +1,141 @@
...
+		if td.Spec.RevisionEnabled {
+			// if any traitDefinition's RevisionEnabled is true
+			// then the component is versioning enabled
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func getComponent(ctx context.Context, client client.Reader, acc v1alpha2.ApplicationConfigurationComponent, namespace string) (*v1alpha2.Component, error) {
```
ryanzhang-oss, line 55: this should be in the common OAM lib

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -65,6 +75,40 @@ func checkRevisionName(appConfig *v1alpha2.ApplicationConfiguration) (bool, stri
...
+	for _, v := range appConfig.Spec.Components {
+		acc := v
+		vEnabled, err := checkComponentVersionEnabled(ctx, client, &acc)
+		if err != nil {
+			return false, fmt.Sprintf(errFmtCheckWorkloadName, err.Error())
+		}
+		if vEnabled {
+			// only check versioning-enabled compoents
+
+			//TODO(roywang) share almost the same render functions with AppConfig controller
```
ryanzhang-oss, line 89: yep, we should not copy the controller logic here

captainroy-hy: Currently, those logics are still private in appConfig controller e.g. Render.go. I'm not sure whether it's allowed for me to make them public... If it's okay, I would like to make them public or move them into common OAM lib (e.g. util/helper.go) and reuse them here.

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -0,0 +1,141 @@
...
+			case intstr.Int:
+				if err := w.SetNumber(path, float64(param.Value.IntVal)); err != nil {
+					return nil, errors.Wrapf(err, errFmtSetParam, param.Name)
+				}
+			}
+		}
+	}
+
+	return &unstructured.Unstructured{Object: w.UnstructuredContent()}, nil
+}
```
wonderflow, line 141: We should refactor the function instead of copy-paste here. Maybe move it to a common helper lib?

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+	return c, nil
+}
+
+type parameter acCtrl.Parameter
+
+func resolveParams(cp []v1alpha2.ComponentParameter, cpv []v1alpha2.ComponentParameterValue) ([]parameter, error) {
+	//TODO(roywang) Maybe other validation will need resolve all params
+	// Just keep it to resolve all params instead of only 'metadata.name'
```
wonderflow, line 81: I think just if there is a parameter contain fieldpath `metadata.name`, then it's invalid in this case. Don't make it too complicated here.

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -65,6 +75,40 @@ func checkRevisionName(appConfig *v1alpha2.ApplicationConfiguration) (bool, stri
...
+			// Can we re-use AppConfig controller functions instead of writing again here?
```
wonderflow, line 90: We MUST re-use these functions instead of re-write.

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -65,6 +75,40 @@ func checkRevisionName(appConfig *v1alpha2.ApplicationConfiguration) (bool, stri
...
 
+// checkWorkloadNameForVersioning check whether versioning-enabled component workload name is empty
+func checkWorkloadNameForVersioning(ctx context.Context, client client.Reader, appConfig *v1alpha2.ApplicationConfiguration) (bool, string) {
+	for _, v := range appConfig.Spec.Components {
+		acc := v
+		vEnabled, err := checkComponentVersionEnabled(ctx, client, &acc)
+		if err != nil {
+			return false, fmt.Sprintf(errFmtCheckWorkloadName, err.Error())
+		}
+		if vEnabled {
```
wonderflow, line 86: if !vEnabled{
  continue
}

less braces is better

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
captainroy-hy: just to stay in the same style as `checkRevisionName(...)(bool,string)`  maybe error is more reasonable. I will fix it.

wonderflow: string is better, as it's not an error, it's reason about why reject.

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
wonderflow: Make them public if there's a need.

captainroy-hy: Got it. I will fix.

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
captainroy-hy: Sure, I will fix it.

captainroy-hy:
@wonderflow @ryanzhang-oss  
 In the new commit,
- move render.go/getComponent and UnpackRevisionData to OAM helper utils
- simplify validation logic, so just reuse OAM helper but no longer rely on render functions of appConfig controller
- modify related unit tests

pkg/webhook/v1alpha2/admit.go
```diff
@@ -8,6 +8,5 @@ import (
 
 // Add will be called in main and register all validation handlers
 func Add(mgr manager.Manager) {
-	server := mgr.GetWebhookServer()
-	applicationconfiguration.Register(server)
+	applicationconfiguration.Register(mgr)
```
wonderflow, line 11: Cool, I didn't know it could work.

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -27,6 +36,8 @@ var appConfigResource = v1alpha2.SchemeGroupVersion.WithResource("applicationcon
 type ValidatingHandler struct {
 	Client client.Client
 
+	AppsClient clientappv1.AppsV1Interface
```
wonderflow, line 39: I wonder if we could use `Client client.Client` instead of create a new one?

captainroy-hy: Yes, that would be more concise, but render.go/getComponent [use clientappv1 to get controllerrevision](https://github.com/crossplane/oam-kubernetes-runtime/blob/ca3bc22d767e0a67da3eeb8bc07360916f544d8d/pkg/controller/v1alpha2/applicationconfiguration/render.go#L254) and I need to reuse that func. So that may be trade-off...

wonderflow: Can you refactor that function?

captainroy-hy: Yes, I will fix it.

captainroy-hy: > Could you help refactor that function?

I create an issue #152 for this and will raise another PR to fix it.

wonderflow: cool, thanks!

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -2,25 +2,42 @@ package applicationconfiguration
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+
```
ryanzhang-oss, line 10: please remove all empty lines and then do a `goimport` to format

pkg/oam/util/helper_test.go
```diff
@@ -880,3 +883,126 @@ var _ = Describe("Test PatchCondition helper utils", func() {
...
+				expectError:        nil,
+			},
+			{
+				caseName:           "get component by componentName",
+				acc:                v1alpha2.ApplicationConfigurationComponent{ComponentName: componentName},
+				expectComponent:    &component2,
+				expectRevisionName: revisionName2,
+				expectError:        nil,
+			},
+		}
```
ryanzhang-oss, line 956: looks like there is no error case?

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -105,7 +106,7 @@ func (c *ComponentHandler) IsRevisionDiff(mt metav1.Object, curComp *v1alpha2.Co
 		c.Logger.Info(fmt.Sprintf("get old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, curComp.Status.LatestRevision.Revision
 	}
-	oldComp, err := UnpackRevisionData(oldRev)
+	oldComp, err := util.UnpackRevisionData(oldRev)
 	if err != nil {
 		c.Logger.Info(fmt.Sprintf("Unmarshal old controllerRevision %s error %v, will create new revision", curComp.Status.LatestRevision.Name, err), "componentName", mt.GetName())
 		return true, oldRev.Revision
```
ryanzhang-oss, line 112: Why not delete the UnpackRevisionData() function at around 121? I noticed that it's test is deleted..

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -0,0 +1,62 @@
...
+			// if any traitDefinition's RevisionEnabled is true
+			// then the component is versioning enabled
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// checkParams will check whethter exist parameter assigning value to workload name
+func checkParams(cp []v1alpha2.ComponentParameter, cpv []v1alpha2.ComponentParameterValue) (bool, string) {
```
ryanzhang-oss, line 46: Does this one have a unit test?

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -65,6 +86,40 @@ func checkRevisionName(appConfig *v1alpha2.ApplicationConfiguration) (bool, stri
 	return true, ""
 }
 
+// checkWorkloadNameForVersioning check whether versioning-enabled component workload name is empty
+func checkWorkloadNameForVersioning(ctx context.Context, client client.Reader, appsClient clientappv1.ControllerRevisionsGetter, appConfig *v1alpha2.ApplicationConfiguration) (bool, string) {
```
ryanzhang-oss, line 90: I strongly recommend testing this individually instead of combining it with the handler test. That is not a good unit test pattern. That test is already way too big, not readable, maintainable.

pkg/webhook/v1alpha2/applicationconfiguration/applicationconfiguration_test.go
```diff
@@ -2,75 +2,284 @@ package applicationconfiguration
...
+		}}})
+	appConfigRevisionName, _ := json.Marshal(v1alpha2.ApplicationConfiguration{
+		ObjectMeta: metav1.ObjectMeta{Namespace: namespace},
+		Spec: v1alpha2.ApplicationConfigurationSpec{Components: []v1alpha2.ApplicationConfigurationComponent{
+			{
+				RevisionName: "r1",
+			},
+		}}})
+	appConfigWithParam, _ := json.Marshal(v1alpha2.ApplicationConfiguration{
+		ObjectMeta: metav1.ObjectMeta{Namespace: namespace},
```
ryanzhang-oss, line 157: try to move all those to their own test

captainroy-hy:
@ryanzhang-oss Got it. I will refine these unit tests and remove the function no longer used.

captainroy-hy:
@ryanzhang-oss @wonderflow I refine the unit tests.

captainroy-hy:
@wonderflow @ryanzhang-oss I refine the unit tests. And it's necessary to let you know that, I remove client-go client in webhook and revert the changes of appConfig controller in this PR because @erdun will remove client-go client in appConfig controller #152 .  To avoid conflicts, I just use newly created GetComponent in OAM helper and leave existing render.go/getComponent **unchanged**.  @erdun could remove old getComponent and UnpackRevisionData in render.go and , then reuse new one in OAM helper latter in his PR.

captainroy-hy:
Sorry for my fault on git ops. This PR is contaminated. I will close and open a new PR for this issue.

Issue #152, opened 2020-07-29
Use one client in appConfig controller to keep code concise


captainroy-hy:
In AppConfig controller [Reconciler](https://github.com/crossplane/oam-kubernetes-runtime/blob/ca3bc22d767e0a67da3eeb8bc07360916f544d8d/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go#L167) and [Component Handler](https://github.com/crossplane/oam-kubernetes-runtime/blob/ca3bc22d767e0a67da3eeb8bc07360916f544d8d/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go#L84), there are two clients instantiated, one is controller-runtime client used to deal with multiple resources, the other one is client-go client only used to handle ControllerRevision resource here. 

Since controller-runtime client also can deal with ControllerRevision as what client-go client does([create](https://github.com/crossplane/oam-kubernetes-runtime/blob/ca3bc22d767e0a67da3eeb8bc07360916f544d8d/pkg/controller/v1alpha2/applicationconfiguration/component.go#L171)&[get](https://github.com/crossplane/oam-kubernetes-runtime/blob/ca3bc22d767e0a67da3eeb8bc07360916f544d8d/pkg/controller/v1alpha2/applicationconfiguration/component.go#L103)), we could just keep controller-runtime client instead of both to keep code concise.

wonderflow:
re-assign to @erdun as he is already working on it #95

captainroy-hy:
Sure, got it.

wonderflow:
fixed by #155

PR #155, opened 2020-07-30
Remove app client

erdun:
Remove app client

See issue #152


pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -168,7 +173,7 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 		Revision: nextRevision,
 		Data:     runtime.RawExtension{Object: curComp},
 	}
-	_, err := c.AppsClient.ControllerRevisions(mt.GetNamespace()).Create(context.Background(), &revision, metav1.CreateOptions{})
+	err := c.Client.Create(context.TODO(), &revision)
```
ryanzhang-oss, line 176: what's the difference between TODO and Background?

erdun: `Background` will never cancel, usually used in top-level, like main functions.
`Todo` used in an unclear context.

PR #159, opened 2020-08-02
add revisions garbage collection mechanism

erdun:
1. add revisions garbage collection mechanism, default revision limit is 50.
2. revision in use won't be deleted

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -16,6 +16,6 @@ package controller
 // Args args used by controller
 type Args struct {
 	// RevisionLimit is the maximum number of revisions that will be maintained.
-	// The default value is 10.
+	// The default value is 50.
 	RevisionLimit int
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1,0 +1,21 @@
+/*
+Copyright 2020 The Crossplane Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+// Args args used by controller
+type Args struct {
+	// RevisionLimit is the maximum number of revisions that will be maintained.
+	// The default value is 10.
+	RevisionLimit int
+}
--- a/pkg/controller/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/component.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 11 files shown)


pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -169,9 +179,74 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+				liveHashes[component.RevisionName] = true
+			}
+		}
+	}
+
+	toKeep := c.RevisionLimit + len(liveHashes)
+	toKill := len(revisions.Items) - toKeep
+	if toKill <= 0 {
+		return nil
+	}
```
wonderflow, line 229: Could you put it into one separated Function and add a unit test for it

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -169,9 +179,74 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+		}
+		// Clean up
+		revisionToClean := revision
+		if err := c.Client.Delete(context.TODO(), &revisionToClean); err != nil {
+			return err
+		}
+		c.Logger.Info(fmt.Sprintf("ControllerRevision %s deleted", revision.Name))
+		toKill--
+	}
+	return nil
```
wonderflow, line 247: Could you split it into two parts(functions), one part is sort and another part is real delete. 

And for delete, we should add a TODO for auditing in the future

pkg/controller/controller.go
```diff
@@ -0,0 +1,21 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+// Args args used by controller
+type Args struct {
+	// RevisionLimit is the maximum number of revisions that will be maintained.
+	// The default value is 10.
```
wonderflow, line 19: Make the default number be 50 will be better I think.

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: 👌

pkg/controller/v1alpha2/applicationconfiguration/component.go
erdun: 👌

pkg/controller/controller.go
erdun: 👌

wonderflow (approved):
LGTM, ping @ryanzhang-oss

Diddaa:
LGTM

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -170,9 +179,86 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+	// get all revisions used and skipped
+	for _, appConfig := range appConfigs {
+		for _, component := range appConfig.Spec.Components {
+			if component.RevisionName != "" {
+				liveHashes[component.RevisionName] = true
+			}
+		}
+	}
+
+	toKeep := revisionLimit + len(liveHashes)
```
ryanzhang-oss, line 205: where is toKeep used other than immediately below?

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -170,9 +179,86 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+	controllerRevisions, toKill, liveHashes := sortedControllerRevision(appConfigs.Items, revisions.Items, c.RevisionLimit)
+	for _, revision := range controllerRevisions {
+		if toKill <= 0 {
+			break
+		}
+		if hash := revision.GetName(); liveHashes[hash] {
+			continue
+		}
+		// Clean up
+		revisionToClean := revision
```
ryanzhang-oss, line 252: same, why creating another variable?

Issue #68, opened 2020-07-02
improve health scope implementation


ryanzhang-oss:
The health scope currently only works with containerizedWorkload. 
- We should make it work with any workload 
-  Add more unit test to bring the code coverage to at least 90%

captainroy-hy:
Maybe I can help on this issue. Could you please assign it to me? @ryanzhang-oss

wonderflow:
I think we could close this issue, thanks to @captainroy-hy

PR #160, opened 2020-08-02
Enhance HealthScope

captainroy-hy:
#68 

- Use adaptable HeathChecker to check different types workload' health status
- Show scope diagnosis info in HealthScope's condition message
- Add built-in checkers for ContainerizedWorkload, Deployment, StatefulSet, and DaemonSet
- Add general check for any type workloads ,e.g OpenKruise CloneSet, except built-in ones, based on `status.availableReplicas` and `status.readyReplicas`
- Add unit tests


pkg/controller/v1alpha2/core/scopes/healthscope/healthscope_controller_test.go
```diff
@@ -0,0 +1,343 @@
...
+			result := reconciler.GetScopeHealthStatus(ctx, &hs)
+			Expect(result).ShouldNot(BeNil())
+			Expect(result.IsHealthy).Should(Equal(true))
+		}
+		//TODO(roywang) golang question:
+		// as shown below
+		// why tc.hsWorkloadRefs passed into It func
+		// shared the same address?
+		// for _, tc := range tests {
+		//
```
wonderflow, line 260: `tc` is a local variable in loop, and the address behind it is an iterator. So when the loop iterated, the value of iterator which is `tc` will change while the address keep the same. Generally, add such line in loop will fix:

```
tc := tc
```

captainroy-hy: Yes, I used `tc:=tc` originally but it didn't work because I put `tc:=tc` inside `It("",func(){tc:=tc ...})`. It should put it before `It` otherwise the scope of `It` will always get the same address of `tc`. Fixed.  Thx!

wonderflow:
ping @artursouza @ryanzhang-oss , could you help review this PR, thanks!

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope_controller.go
```diff
@@ -36,20 +40,20 @@ import (
...
-	shortWait        = 30 * time.Second
-	longWait         = 1 * time.Minute
+	// shortWait        = 30 * time.Second
+	longWait = 1 * time.Minute
 )
 
 // Reconcile error strings.
 const (
-	errGetHealthScope          = "cannot get health scope"
+	errGetHealthScope          = "cannotrget health scope"
```
artursouza, line 49: typo.

artursouza:
BTW, I am not against this PR but this cannot be extended to future workloads.

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -19,133 +19,244 @@ package healthscope
...
+	for _, fp := range generalReplicaCheckFiled {
+		if value, err := pavedV.GetNumber(fp); err == nil {
+			// just check ready/available replica exists
+			if value != 0 {
+				return &HealthCondition{
+					Target:    ref,
+					IsHealthy: true,
+				}
+			}
+			//TODO(roywang) does every workload have status?
```
ryanzhang-oss, line 250: not really

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -19,133 +19,244 @@ package healthscope
...
-		workloadObject := unstructured.Unstructured{}
-		workloadObject.SetAPIVersion(workloadRef.APIVersion)
-		workloadObject.SetKind(workloadRef.Kind)
-		workloadObjectRef := types.NamespacedName{Namespace: healthScope.GetNamespace(), Name: workloadRef.Name}
-		if err := client.Get(ctxWithTimeout, workloadObjectRef, &workloadObject); err != nil {
-			return errors.Wrapf(err, errNoWorkload, workloadRef.Name)
-		}
+var (
+	// for general check on worload's replicas
+	generalReplicaCheckFiled = []string{
```
ryanzhang-oss, line 53: not all resources have these two fields

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -19,133 +19,244 @@ package healthscope
...
 	}
+	ref.UID = o.GetUID()
+	pavedV := fieldpath.Pave(o.UnstructuredContent())
 
-	if deployment.Status.ReadyReplicas == 0 {
-		return fmt.Errorf(errDeploymentUnavailable, ref.APIVersion, ref.Kind, ref.Name)
+	for _, fp := range generalReplicaCheckFiled {
+		if value, err := pavedV.GetNumber(fp); err == nil {
+			// just check ready/available replica exists
+			if value != 0 {
```
ryanzhang-oss, line 244: just one replica exists doesn't mean it's healthy...

captainroy-hy:
> This works for the known K8s resources but this is not extensible. We had discussed about having a HealthCheckTrait to decouple health check from health scope. HealthScope is the aggregation of health statuses while the health status itself is the output of one kind of health check.

@artursouza  Yes, I agree that implementation in this PR is not extensible, and `HealthCheckTrait` is a good idea for adopting any kind of health check while just leave `HealthScope` playing a role as an aggregation of health statuses. 
However, is it okay to keep several built-in health checkers in `HealthScope` for core workloads and native K8s resources, just like what we handled in this PR? And with further enhancement, we can make `HealthCheckTrait` possible to override built-in checkers according to users' config. Then, users can just put Component into HealthScope without assigning additional `HealthCheckTrait` every time, unless they really need custom `HealthCheckTrait`.

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
captainroy-hy: Yes, it seems such kind of general checking rule doesn't make sense... Maybe I should remove `generalReplicaCheck` and leave it to `HealthCheckTrait` to satisfy diverse health checking needs.

wonderflow:
@captainroy-hy  Yes, I think this PR is good, and we could support `HealthCheckTrait` in the following PRs.

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -19,133 +19,244 @@ package healthscope
...
-	errResourceNotFound      = "could not retrieve resource %q %q %q"
-	errDeploymentUnavailable = "no ready instance found in %q %q %q"
+	errFmtUnsupportWorkload   = "APIVersion %v Kind %v workload is not supportted by HealthScope"
+	errHealthCheck            = "error occurs in health check"
+	errUnhealthyChildResource = "unhealthy child resource exists"
+	errFmtResourceNotReady    = "resource not ready, resource status: %+v"
 
 	defaultTimeout = 10 * time.Second
+
+	kindContainerizedWorkload = "ContainerizedWorkload"
```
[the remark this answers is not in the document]
captainroy-hy: Sure, I will fix it.

captainroy-hy:
Hi @ryanzhang-oss  I remove general health check and fix hardcode kind name issue. Could you please help review this PR, thanks!

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -19,133 +19,198 @@ package healthscope
...
+	deployment.SetGroupVersionKind(apps.SchemeGroupVersion.WithKind(kindDeployment))
 	deploymentRef := types.NamespacedName{Namespace: namespace, Name: ref.Name}
 	if err := client.Get(ctx, deploymentRef, &deployment); err != nil {
-		return errors.Wrapf(err, errResourceNotFound, ref.APIVersion, ref.Kind, ref.Name)
+		r.Diagnosis = errors.Wrap(err, errHealthCheck).Error()
+		return r
 	}
+	r.Target.UID = deployment.GetUID()
 
 	if deployment.Status.ReadyReplicas == 0 {
```
ryanzhang-oss, line 156: I may be missing something here, why is that "ReadyReplicas != 0" means healthy?

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -19,133 +19,198 @@ package healthscope
...
+	statefulset.APIVersion = ref.APIVersion
+	statefulset.Kind = ref.Kind
+	nk := types.NamespacedName{Namespace: namespace, Name: ref.Name}
+	if err := client.Get(ctx, nk, &statefulset); err != nil {
+		r.Diagnosis = errors.Wrap(err, errHealthCheck).Error()
+		return r
+	}
+	r.Target.UID = statefulset.GetUID()
+
+	if statefulset.Status.ReadyReplicas == 0 {
```
ryanzhang-oss, line 183: same here, why is that "ReadyReplicas != 0" means healthy?

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
captainroy-hy: Actually I just stay in line with [current logic on health checking](https://github.com/crossplane/oam-kubernetes-runtime/blob/bc0db9fd48f9d577c6bb4cf6f715cffded6416ed/pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go#L147) .And a more rigorous check is `x.Status.ReadyReplicas == x.Spec.Replicas`

wonderflow: I think we could fix it in the following PRs

wonderflow (approved):
I think this PR is good, let me merge it, and fix other things in the following PRs.

PR #162, opened 2020-08-06
allow PreserveUnknownFields for extention field of definition

wonderflow:
Add this tag, so it won't prevent unknow fields in extention.


hongchaodeng:
good catch!

PR #163, opened 2020-08-06
add finalizer in AppConfig controller for workload scope

captainroy-hy:
- Use finalizer to enable pre-delete hooks on AppConfig
- WorkloadScopeFinalizer: Remove scopes' workload refs owned by AppConfig being deleted
- Add unit tests


pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -190,6 +193,7 @@ func NewReconciler(m ctrl.Manager, o ...ReconcilerOption) *OAMApplicationReconci
 
 // Reconcile an OAM ApplicationConfigurations by rendering and instantiating its
 // Components and Traits.
+// nolint:gocyclo
```
wonderflow, line 196: maybe we should keep the go-lint check and make the function clean. We can separate that logic out of the function

captainroy-hy: I try to separate the logic of finalizers from Reconcile() func, but it's still in need of one `if err!=nil` in Reconcile func and that makes cyclomatic complexiy into 13 while gocyclo max is 12...  Is it okay for me to combine Render&Apply logic into one func and separate it from Reconcile? Or we can set the threshold of gocyclo higher,e.g. `gocyclo -over 15`?

wonderflow: 15 is OK I think

pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -131,6 +159,20 @@ func (a *workloads) dereferenceScope(ctx context.Context, namespace string, stat
 	return nil
 }
 
+// dereferenceAllScope dereferences workloads owned by the appConfig being deleted from the scopes they belong to.
+func (a *workloads) dereferenceAllScopes(ctx context.Context, namespace string, status []v1alpha2.WorkloadStatus) error {
+	for _, st := range status {
+		toBeDeferenced := st.Scopes
+		for _, s := range toBeDeferenced {
+			if err := a.applyScopeRemoval(ctx, namespace, st, s); err != nil {
```
ryanzhang-oss, line 167: why pass in st again after it's already used here? Only the workloadRef from status is needed

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
wonderflow, line 196: we should remove this line since the `min-complexity: 15` has changed

captainroy-hy:
> finalizer will block delete if we are not handle it right, so I think we must write a e2e test for this case.
> 
> Cases must include:
> 
> 1. appconfig without scopes will not have finalizer added and can work smoothly
> 2. appcofig with scopes have finalizer and can be deleted normally

Yes, got it. I will add e2e test for these cases.

wonderflow (approved):
excellent job!

Issue #167, opened 2020-08-07
Resource Dependency to support appending array of unstructured types


hongchaodeng:
Have talked offline to some users and get the following feedback:

In the [Cloud logging setup](https://help.aliyun.com/document_detail/131598.html?spm=a2c4g.11174283.6.1009.103d2ceeD8wee2), it requires to pass `[]v1.EnvVar`, `[]v1.Volume`, `[]v1.VolumeMount` as the parameters:

```yaml
      containers:
      - image: alpine
        imagePullPolicy: Always
        args:
        - ping
        - 127.0.XX.XX
        name: alpine
        env:
        - name: aliyun_logs_project
          value: k8s-xxx
        - name: aliyun_logs_machinegroup
          value: k8s-group-app-alpine
        - name: aliyun_logs_log-stdout
          value: stdout
        - name: aliyun_logs_test-file
          value: /log/*.log
        - name: aliyun_logs_test-stdout_ttl
          value: "7"
        - name: aliyun_logs_test-stdout_shard
          value: "2"
        volumeMounts:
        - name: volume-sls
          mountPath: /log
      volumes:
      - name: volumn-sls
        emptyDir: {}
```

However, the current implementation doesn't support such array types yet.

PR #169, opened 2020-08-09
dependency: support appending array in data input

hongchaodeng:
fix https://github.com/crossplane/oam-kubernetes-runtime/issues/167

With this PR, the dependency engine can now take array in data input as well:

<details>
<summary>Click to toggle the example AppConfig:</summary>

```yaml
kind: Component
metadata:
  name: source
spec:
  workload:
    apiVersion: example.com/v1
    kind: Foo
    metadata:
      name: source
    status:
      key:
        - name: a
          value: aa
        - name: b
          value: bb
---
kind: Component
metadata:
  name: sink
spec:
  workload:
    apiVersion: example.com/v1
    kind: Foo
    metadata:
      name: sink
    spec:
      key:
        - name: exist
          value: existtt
---
metadata:
  name: example-appconfig
spec:
  components:
    - componentName: source
      dataOutputs:
        - name: example-key
          fieldPath: "status.key"
    - componentName: sink
      dataInputs:
        - valueFrom:
            dataOutputName: example-key
          toFieldPaths:
            - "spec.key"
```
</details>

<details>
<summary>Click to toggle the result</summary>

```yaml
apiVersion: example.com/v1
kind: Foo
metadata:
  name: sink
  ownerReferences:
  - apiVersion: core.oam.dev/v1alpha2
    blockOwnerDeletion: true
    controller: true
    name: example-appconfig
spec:
  key:
  - name: exist
    value: existtt
  - name: a
    value: aa
  - name: b
    value: bb
```
</details>

I will test with user scenarios first before removing WIP.

### DIFF
--- a/apis/core/v1alpha2/core_types.go
+++ b/apis/core/v1alpha2/core_types.go
@@ -461,6 +461,7 @@ type DataOutput struct {
 }
 
 // DataInput specifies a data input sink to an object.
+// If input is array, it will be appended to the target field paths.
 type DataInput struct {
 	// ValueFrom specifies the value source.
 	ValueFrom DataInputValueFrom `json:"valueFrom,omitempty"`
--- a/pkg/controller/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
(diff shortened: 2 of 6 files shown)


apis/core/v1alpha2/core_types.go
```diff
@@ -461,6 +461,7 @@ type DataOutput struct {
 }
 
 // DataInput specifies a data input sink to an object.
+// If input is array, it will be appended to the target field paths.
```
wonderflow, line 464: what if some scenario need replace the whole array instead of append in the future？

hongchaodeng: In that case we need to specify the patch behavior in data input. But until now I don't any use case for it.

XianB:
I have tested the code with my use case and it works great!

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -440,18 +441,35 @@ func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInp
...
+			l := raw.([]interface{})
+			l = append(l, val.([]interface{})...)
+			toSet = l
+		}
+
+		err := paved.SetValue(fp, toSet)
 		if err != nil {
-			return fmt.Errorf("paved.SetString() failed: %w", err)
+			return errors.Wrap(err, "paved.SetValue() failed")
 		}
```
wonderflow, line 467: Please add a test case for it

Issue #110, opened 2020-07-21
Multiple revision of one component in the AppConfig


wonderflow:
```
apiVersion: core.oam.dev/v1alpha2
kind: ApplicationConfiguration
metadata:
  name: my-app-deployment
  annotations:
    version: v1.0.0
    description: "Description of this deployment"
spec:
  components:
    - revisionName: example-component-1
      traits:
        - apiversion:core.oam.dev/v1alpha2
          kind: ManualScaler
          spec:
            replicaCount: 3
    - revisionName: example-component-2
      traits:
        - apiversion:core.oam.dev/v1alpha2
          kind: ManualScaler
          spec:
            replicaCount: 2
```

In this example, revision  ` example-component-1` and `example-component-2` are of the same component.

Add this case into revision proposal, and make sure this case work.

captainroy-hy:
It comes across with below confusing scenario. 
Here's the schematic of a Component
```yaml
apiVersion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: example-component
spec:
  workload:
    apiVersion: core.oam.dev/v1alpha2
    kind: ContainerizedWorkload
    spec:
      containers:
        - name: wordpress
          image: wordpress:php7.2
  parameters:
    - name: instance-name
      required: true
      fieldPaths:
        - metadata.name # this param allows to override the name of workload obj
```
And the appConfig contains two revision of one component but they are assigned the same workload name.
```yaml
apiVersion: core.oam.dev/v1alpha2
kind: ApplicationConfiguration
metadata:
  name: example-appconfig
spec:
  components:
    - revisionName: example-component-bs48sm18d3b8dpjvdab0
      parameterValues:
        - name: instance-name
          value: same-workload-name
    - revisionName: example-component-bs4l2f98d3b8dpjvdabg
      parameterValues:
        - name: instance-name
          value: same-workload-name
```
The deployment result is that there's only one ContainerizedWorkload object created, and in each round of reconciling, one more pod is being created but terminated immediately.
```shell
$ kubectl get pods
NAME                                  READY   STATUS        RESTARTS   AGE
same-workload-name-7b9ff9bb67-ztcsz   0/1     Terminating   0          15s
same-workload-name-854c9cf678-94wv6   1/1     Running       0          2m24s
```
---
[...]

PR #171, opened 2020-08-09
Component versioning fix and e2e tests

### LINKED ISSUE #170: [BUG] missing namespace setting when create ControllerRevision
**Describe the bug**

ControllerRevision cannot be created since namespace value is not assigned. That breaks Component versioning mechanism. PR #155 
`error cannot create a controllerrevision: an empty namespace may not be set during creation`

captainroy-hy:
- Allow multiple revisions of one component in the appConfig. #110 
Change: set workload instance name as specified revisionName, if revisionName is provided.
- Fix #170 
- Add e2e-test for Component Versioning mechanism
- Add and modify unit tests.

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/component.go
@@ -144,7 +144,8 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 	// set annotation to component
 	revision := appsv1.ControllerRevision{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: revisionName,
+			Name:      revisionName,
+			Namespace: curComp.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					APIVersion: v1alpha2.SchemeGroupVersion.String(),
--- a/pkg/controller/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render.go
@@ -149,7 +149,7 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 		traits = append(traits, &Trait{Object: *t})
 		traitDefs = append(traitDefs, *traitDef)
 	}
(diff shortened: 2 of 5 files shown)


pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -144,7 +144,8 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 	// set annotation to component
 	revision := appsv1.ControllerRevision{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: revisionName,
+			Name:      revisionName,
+			Namespace: curComp.Namespace,
```
wonderflow, line 148: could you help split the namespace in a separated PR as a BUG fix, thanks.

design/one-pager-component-mutable-and-versioning.md
```diff
@@ -359,6 +359,34 @@ spec:
...
+        - apiversion:core.oam.dev/v1alpha2
+          kind: ManualScaler
+          spec:
+            replicaCount: 3
+    - revisionName: example-component-2
+      traits:
+        - apiversion:core.oam.dev/v1alpha2
+          kind: ManualScaler
+          spec:
+            replicaCount: 2
```
wonderflow, line 387: We are considering it's an anti-pattern, as we could use two different components to do the same thing. Using different revisions of one component  seems don't bring any extra bonus.

pkg/controller/v1alpha2/applicationconfiguration/component.go
captainroy-hy: Sure, I will fix it and modify unit tests in another PR.

design/one-pager-component-mutable-and-versioning.md
captainroy-hy: Okay, before the proposal is ready, let's suspend this PR. But I will separate e2e test covering current versioning mechanism into another PR as well as the fix of #170.

wonderflow: Cool, we can hold this PR until we have a conclusion on that issue

PR #183, opened 2020-08-26
patch the status instead of update

ryanzhang-oss:
Patch the AC status instead of update at the end of reconcile loop


wonderflow (approved):
LGTM

Diddaa:
LGTM

PR #189, opened 2020-08-29
add labels for workload to record component info

captainroy-hy:
- add labels for workloads to record component info 
- add e2e test & unit test


wonderflow (approved):
LGTM , ping @ryanzhang-oss @zzxwill

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -130,6 +131,13 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 		return nil, errors.Wrapf(err, errFmtRenderWorkload, acc.ComponentName)
 	}
 
+	labels := map[string]string{
+		oam.LabelAppName:              ac.Name,
+		oam.LabelAppComponent:         acc.ComponentName,
+		oam.LabelAppComponentRevision: componentRevisionName,
+	}
+	w.SetLabels(labels)
```
wonderflow, line 139: please make sure we won't override the existing labels.

captainroy-hy: sure, I will use merge instead of override here.

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -469,6 +477,17 @@ func fillValue(obj *unstructured.Unstructured, fs []string, val interface{}) err
 	return nil
 }
 
+func addWorkloadLabels(w *unstructured.Unstructured, labels map[string]string) {
```
ryanzhang-oss, line 480: can you move this to util and change the name from "addWorkloadLabels" to "addLabels" as this is very generic. Also, I am not sure if this already exists somewhere in the library

captainroy-hy: Yes, I will check and move it.

ryanzhang-oss:
Good job

PR #191, opened 2020-09-01
use human readable revision name instead of hash

wonderflow:


wonderflow:
@ryanzhang-oss  yes, they are all string

PR #204, opened 2020-09-09
fix: change resource type appconfigs to appconfig

dakele123:
Signed-off-by: dakele <wmxy123@yeah.net>

kubernetes say it doesn't have appconfigs.


wonderflow (approved):
LGTM thanks

PR #205, opened 2020-09-13
feat(component):Add observedGeneration to represent Component

### LINKED ISSUE #91: add observedGeneration to represent Component and AppConfig has reconciled
`observedGeneration` field of status in K8s object is widely used in K8s built-in resource. It reflects the generation of the most recently observed Object.

With the help of `observedGeneration` field, user can recognize whether a CR is reconciled by controller.

A workflow will like below:

```
apiversion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: c1
  generation: 1
spec:
  workload:
    image: nginx:v1
    ...
```

After the controller reconciled, the `observedGeneration` will appear and align with `generation` in metadata:

```
apiversion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: c1
  generation: 1
spec:
  workload:
    image: nginx:v1
    ...
status:
  observedGeneration: 1
```

Later, we have another change, it will be:

```
apiversion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: c1
  generation: 2
spec:
  workload:
    image: nginx:v2
    ...
status:
  observedGeneration: 1
```

And after reconciled, it will be:

```
apiversion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: c1
  generation: 2
spec:
  workload:
    image: nginx:v2
    ...
status:
  observedGeneration: 2
```

Also, CR of AppConfig also need to add this field.

Ghostbaby:
Add observedGeneration to represent Component

Fix #91

Signed-off-by: zhuhuijun <zhuhuijunzhj@gmail.com>


apis/core/v1alpha2/core_types.go
```diff
@@ -221,6 +221,10 @@ type ComponentSpec struct {
 
 // A ComponentStatus represents the observed state of a Component.
 type ComponentStatus struct {
+	// The generation observed by the component controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration"`
```
wonderflow, line 226: We also need ObservedGeneration in status of AppConfig.

wonderflow (commented):
Could you please add a test for this feature

Ghostbaby:
copy

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -347,6 +347,10 @@ func (r *OAMApplicationReconciler) updateStatus(ctx context.Context, ac *v1alpha
 	}
 	ac.Status.Workloads = append(ac.Status.Workloads, revisionStatus...)
 
+	if ac.Status.ObservedGeneration != ac.Generation {
+		ac.Status.ObservedGeneration = ac.Generation
```
wonderflow, line 351: please make sure the ObservedGeneration is configured in every logic branch of the reconcile function.

Now it's only configured when the reconcile is succeed, but when one reconcile fail, users will also hope to see this CR is observed

wonderflow (changes requested):

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration_test.go
```diff
@@ -22,13 +22,18 @@ import (
 	"testing"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+
```
wonderflow, line 28: nit, please put them into their group

wonderflow (approved):
Generally LGTM, please fix the conflict， thanks!

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -228,6 +228,7 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 
 	// execute the posthooks at the end no matter what
 	defer func() {
+		updateObservedGeneration(ac)
```
hongchaodeng, line 231: An afterthought: I don't think this is the right place to update ObservedGeneration. Because this will always update even if the reconciliation failed.

Note that I will be fixing this shortly.

@wonderflow @Ghostbaby

wonderflow: Reconcile failed also need to add observed

PR #206, opened 2020-09-13
Fix helm chart template format issues

SataQiu:
`nodeSelector` should be a member of `spec.template.spec`, not `spec.template`.

The error reported by Helm is:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template): unknown field "nodeSelector" in io.k8s.api.core.v1.PodTemplateSpec
```

`affinity` and `tolerations` have similar problems :(


wonderflow (approved):
LGTM thanks

Issue #207, opened 2020-09-16
[BUG] Can't find workloadDefinition is name was not match pattern <puraral-kind>.<resource-group>


wonderflow:
**Describe the bug**

In #195 and #198 , we have created some workload definition have custom name.

When working with Definition like [this](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/examples/typed-component/definitions.yaml#L4). The controller can't work, it says can't find workloadDefinition as we don't follow the old convention that definition name will always equal to CRD name which match the pattern `<puraral-kind>.<resource-group>`

zzxwill:
This happens when WorkloadDefinition's name doesn't follow the convention of `<puraral-kind>.<resource-group>`. For example.
- WorkloadDefinition with name `web-service`
```
apiVersion: core.oam.dev/v1alpha2
kind: WorkloadDefinition
metadata:
  annotations:
    definition.oam.dev/name: web-service
  name: web-service
spec:
  definitionRef:
    name: containerizedworkloads.core.oam.dev
  childResourceKinds:
    - apiVersion: apps/v1
      kind: Deployment
    - apiVersion: v1
      kind: Service
```

- component
```
apiVersion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: example-component
spec:
  workload:
    apiVersion: core.oam.dev/v1alpha2
    kind: ContainerizedWorkload
    metadata:
      annotations:
        definition.oam.dev/name: web-service
    spec:
      containers:
        - name: wordpress
          image: wordpress:4.6.1-apache
          ports:
            - containerPort: 80
              name: wordpress
          env:
            - name: TEST_ENV
              value: test
  parameters:
    - name: instance-name
      fieldPaths:
        - metadata.name
    - name: image
      fieldPaths:
        - spec.containers[0].image
```

- ApplicationConfiguration
```
apiVersion: core.oam.dev/v1alpha2
kind: ApplicationConfiguration
metadata:
  name: example-appconfig
spec:
  components:
    - componentName: example-component
      parameterValues:
        - name: image
          value: wordpress:php7.2
      traits:
        - trait:
            apiVersion: core.oam.dev/v1alpha2
            kind: ManualScalerTrait
            metadata:
              name: example-appconfig-trait
            spec:
              replicaCount: 3
```

After these manifest are applied, it will hit the issue as below.
```
[...]

Issue #195, opened 2020-09-03
[Feature]OAM Runtime should allow different workloadDefinitions to point to the same workload Schema


ryanzhang-oss:
**Is your feature request related to a problem? Please describe.**
Currently,  we only allow one workloadDefinition to point to the one workload Schema by tying its name to the schema name of the workload. This leads to the problem of not able to express different types of workloads with the same schema. 

**Describe the solution you'd like**
We will decouple the name of the workloadDefintion from the name of workload schema. In order to achieve this, we will propose to add an optional "type" field in the Component to point to the name of the "workloadDefinition" it is referring to. This is fully backward compatible with the current schema. Here is an example.

```yaml
apiVersion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: frontend
spec:
  workload:
    type: Server
    spec:
      image: nginx
      port: 80
```
 

**Describe alternatives you've considered**
Add a workloadDefinition controller that watches all the resources of type "CRD" and add an owner reference to itself when it detects a new resource of type "CRD" is created. This is a pretty heavy process and can lead to user confusion.

resouer:
The `workload` hierarchy on top of `spec` seems useless. Can we avoid that?

For example:

```yaml
apiVersion: core.oam.dev/v1alpha2
kind: Component
metadata:
  name: frontend
spec:
  workload:
    type: Server
    spec:
      image: nginx
      port: 80
```

For backward compatible implementation, the go types in my mind will be:

```go
type ComponentSpec struct {
	Workload WorkloadTemplate
	Parameters []ComponentParameter
}
type WorkloadTemplate struct {
	Type string
	Spec runtime.RawExtension
        runtime.RawExtension // for GVK object, this field is exclusive to Type
}
```

resouer:
This also makes it right that `workload` section is a template of workload instances.

ryanzhang-oss:
I am not sure if this is compatible with the current structure. as the workload itself is already declared as a RawExtention. It might be doable but pretty tricky and not sure what does that buy. The below structure is fully backward compatible and easy to implenent.

type ComponentSpec struct {
        Type               String
	Workload        runtime.RawExtension 
	Parameters     []ComponentParameter 
}

ryanzhang-oss:
[...]

PR #208, opened 2020-09-16
Fix `could not find WorkloadDefinition` issue

zzxwill:
When WorkloadDefinition doesn't follow the pattern
<puraral-kind>.<resource-group> or there are different
workloadDefinitions pointing to the same workload Schema,
it will hit the issue of could not finding workloaddefinition.
Fix issue #207, feature #195.

Signed-off-by: zzxwill <zzxwill@gmail.com>


zzxwill:
I will add unit-test and e2e-test tomorrow to make the PR ready for review.

https://github.com/crossplane/oam-kubernetes-runtime/issues/207#issuecomment-693524391 is an introduction and comparison with the solution of #195.

pkg/oam/util/helper.go
```diff
@@ -226,7 +228,11 @@ func PassLabelAndAnnotation(parentObj oam.Object, childObj labelAnnotationObject
 
 // GetCRDName return the CRD name of any resources
 // the format of the CRD of a resource is <kind purals>.<group>
+// Now the CRD name of a resource could also be defined as `metadata.annotations.definition.oam.dev/name`
```
wonderflow, line 231:
```suggestion
// Now the CRD name of a resource could also be defined as `definition.oam.dev/name` in `metadata.annotations`
```

pkg/oam/util/helper.go
```diff
@@ -226,7 +228,11 @@ func PassLabelAndAnnotation(parentObj oam.Object, childObj labelAnnotationObject
...
 func GetCRDName(u *unstructured.Unstructured) string {
+	if crdName, ok := u.GetAnnotations()[DefinitionAnnotation]; ok {
```
wonderflow, line 233: using index like this will easily cause panic if the map returned is nil

pkg/oam/util/helper.go
```diff
@@ -39,6 +39,8 @@ var (
 const (
 	// TraitPrefixKey is prefix of trait name
 	TraitPrefixKey = "trait"
+	// DefinitionAnnotation is the annotation of filed `metadata.annotations` for WorkloadDefinition and Workload
```
wonderflow, line 42:
```suggestion
	// DefinitionAnnotation is one automatically generated annotation of workload and trait to indicate which workloadDefinition/ traitDefinition it is generated
```

wonderflow (changes requested):

pkg/oam/util/helper.go
```diff
@@ -39,6 +39,9 @@ var (
 const (
 	// TraitPrefixKey is prefix of trait name
 	TraitPrefixKey = "trait"
+	// DefinitionAnnotation is one automatically generated annotation of workload and trait to indicate which
+	//workloadDefinition/traitDefinition it is generated by
+	DefinitionAnnotation = "definition.oam.dev/name"
```
wonderflow, line 44: Maybe put this annotation const to types.go is better

pkg/oam/util/helper.go
```diff
@@ -226,7 +226,13 @@ func PassLabelAndAnnotation(parentObj oam.Object, childObj labelAnnotationObject
 
 // GetCRDName return the CRD name of any resources
 // the format of the CRD of a resource is <kind purals>.<group>
+// Now the CRD name of a resource could also be defined as `definition.oam.dev/name` in `metadata.annotations`
 func GetCRDName(u *unstructured.Unstructured) string {
```
wonderflow, line 230: Please also refactor the function Name, it should not be `GetCRDName ` , name it `GetDefinitionName` seems better

wonderflow (approved):
LGTM, In this PR, we add an annotation into workload to show which Definition it is generated from.

pkg/webhook/v1alpha2/component/mutating_handler.go
```diff
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam"
```
ryanzhang-oss, line 25: please move this with the local

pkg/webhook/v1alpha2/component/mutating_handler.go
```diff
@@ -125,7 +127,9 @@ func (h *MutatingHandler) Mutate(obj *v1alpha2.Component) error {
 		// copy namespace/label/annotation to the workload and add workloadType label
 		workload.SetNamespace(obj.GetNamespace())
 		workload.SetLabels(util.MergeMap(obj.GetLabels(), map[string]string{WorkloadTypeLabel: workloadType}))
-		workload.SetAnnotations(obj.GetAnnotations())
+		// Add another annotation DefinitionAnnotation which can mark the name of WorkloadDefinition
+		workload.SetAnnotations(util.MergeMap(obj.GetAnnotations(), map[string]string{oam.DefinitionAnnotation: workloadType}))
+		mutatelog.Info("Set annotation definition.oam.dev/name for workload", "annotation value", workloadType)
```
ryanzhang-oss, line 132: this seems to be a debugging log for the testing purpose only

pkg/webhook/v1alpha2/component/mutating_handler.go
```diff
@@ -125,7 +127,9 @@ func (h *MutatingHandler) Mutate(obj *v1alpha2.Component) error {
 		// copy namespace/label/annotation to the workload and add workloadType label
 		workload.SetNamespace(obj.GetNamespace())
 		workload.SetLabels(util.MergeMap(obj.GetLabels(), map[string]string{WorkloadTypeLabel: workloadType}))
-		workload.SetAnnotations(obj.GetAnnotations())
+		// Add another annotation DefinitionAnnotation which can mark the name of WorkloadDefinition
+		workload.SetAnnotations(util.MergeMap(obj.GetAnnotations(), map[string]string{oam.DefinitionAnnotation: workloadType}))
```
ryanzhang-oss, line 131: there is already a label to indicate the workloadType of the component. Where is this annotation with the same information used?

pkg/webhook/v1alpha2/component/component_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam"
```
ryanzhang-oss, line 7: move this to its group

pkg/oam/util/helper.go
```diff
@@ -224,9 +224,15 @@ func PassLabelAndAnnotation(parentObj oam.Object, childObj labelAnnotationObject
...
 
-// GetCRDName return the CRD name of any resources
-// the format of the CRD of a resource is <kind purals>.<group>
-func GetCRDName(u *unstructured.Unstructured) string {
+// GetDefinitionName return the Definition name of any resources
+// the format of the definition of a resource is <kind plurals>.<group>
+// Now the definition name of a resource could also be defined as `definition.oam.dev/name` in `metadata.annotations`
+func GetDefinitionName(u *unstructured.Unstructured) string {
+	if annotations := u.GetAnnotations(); annotations != nil {
+		if crdName, ok := annotations[oam.DefinitionAnnotation]; ok {
```
ryanzhang-oss, line 232: why not use the existing label and add this extra annotation?

pkg/webhook/v1alpha2/component/mutating_handler.go
wonderflow: Oh,yes, using the existing WorkloadTypeLabel is better.

pkg/oam/util/helper.go
zzxwill: Will use the label, thanks.

pkg/webhook/v1alpha2/component/mutating_handler.go
zzxwill: Webhook is a little bit for local debug, so I write the log to make sure it takes effect. As you recommended to use the existing label below, I will remove the annotation setting and the log. Thanks.

pkg/webhook/v1alpha2/component/mutating_handler.go
zzxwill: Will use the label, instead.

pkg/oam/util/helper.go
```diff
@@ -224,9 +224,15 @@ func PassLabelAndAnnotation(parentObj oam.Object, childObj labelAnnotationObject
...
 }
 
-// GetCRDName return the CRD name of any resources
-// the format of the CRD of a resource is <kind purals>.<group>
-func GetCRDName(u *unstructured.Unstructured) string {
+// GetDefinitionName return the Definition name of any resources
+// the format of the definition of a resource is <kind plurals>.<group>
+// Now the definition name of a resource could also be defined as `definition.oam.dev/name` in `metadata.annotations`
+func GetDefinitionName(u *unstructured.Unstructured) string {
+	if labels := u.GetAnnotations(); labels != nil {
```
wonderflow, line 231: We need to get labels now, not annotations

wonderflow: If tests still passed with this obviously bug, you need to write more cases I think.

pkg/webhook/v1alpha2/component/component_test.go
```diff
@@ -121,6 +123,7 @@ var _ = Describe("Component Admission controller Test", func() {
 			mutatedWorkload := baseWorkload.DeepCopy()
 			mutatedWorkload.SetNamespace(component.GetNamespace())
 			mutatedWorkload.SetLabels(util.MergeMap(label, map[string]string{WorkloadTypeLabel: workloadTypeName}))
+			mutatedWorkload.SetAnnotations(map[string]string{oam.DefinitionAnnotation: ""})
```
wonderflow, line 126: If we use label, why still DefinitionAnnotation here?

pkg/oam/util/helper.go
zzxwill: > If tests still passed with this obviously bug, you need to write more cases I think.

Sorry, the code was automatically reverted from Disk instead of Memory by Goland. My bad, fixed. Thanks.

PR #212, opened 2020-09-19
fix status automatically removed bug

ryanzhang-oss:
The status fields in the workload disappear after a reconcile loop.  The culprit is that the controller runtime merge function doesn't know how to patch a simple Json object... The solution is to manually patch it on the clientside (lame)

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
@@ -208,6 +208,7 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 	if err := r.client.Get(ctx, req.NamespacedName, ac); err != nil {
 		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errGetAppConfig)
 	}
+	acPatch := ac.DeepCopy()
 
 	if ac.ObjectMeta.DeletionTimestamp.IsZero() {
 		if registerFinalizers(ac) {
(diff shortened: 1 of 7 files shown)


hongchaodeng (approved):
code lgtm
Might need add some UT to pass codecov.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -346,10 +345,36 @@ func (r *OAMApplicationReconciler) updateStatus(ctx context.Context, ac *v1alpha
 		}
 	}
 	ac.Status.Workloads = append(ac.Status.Workloads, revisionStatus...)
-
+	// patch the extra fields in the status that is wiped by the Status() function
```
hongchaodeng, line 348: `Status()` is not specific enough to pinpoint the code location.
Better point out its parent function/object as well.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -346,10 +345,36 @@ func (r *OAMApplicationReconciler) updateStatus(ctx context.Context, ac *v1alpha
...
+	patchExtraStatusField(&ac.Status, acPatch.Status)
 	ac.SetConditions(v1alpha1.ReconcileSuccess())
 }
 
+func patchExtraStatusField(acStatus *v1alpha2.ApplicationConfigurationStatus, acPatchStatus v1alpha2.ApplicationConfigurationStatus) {
```
wonderflow, line 353: This function has  a problem, workload status which should be removed will always leave here now.

wonderflow (approved):
LGTM

wonderflow:
The Coverage seems to have some problem, merging

PR #214, opened 2020-09-22
add health check support for containerized.standard.oam.dev

captainroy-hy:
- add health check support for containerized.standard.oam.dev
- add unit tests

@wonderflow I'm not sure whether it's okay to import `kubevela` module as dependency, so I use unstructured type in the health checker of `containerized.standard.oam.dev`.


wonderflow (approved):
Generally LGTM

PR #220, opened 2020-09-25
separate history workloads out from workloadStatus

### LINKED ISSUE #217: [Feature] history workload in status should be in a separated list not align with current workload
Currently, it's confusing.

Status in red block should in a separated field like `historyWorkloads`.

image

wonderflow:
fixes #217


ryanzhang-oss (approved):
LGTM, just some wording issue

apis/core/v1alpha2/core_types.go
```diff
@@ -389,6 +386,15 @@ type WorkloadStatus struct {
 	Scopes []WorkloadScope `json:"scopes,omitempty"`
 }
 
+// HistoryWorkload record history but still working revision workload.
```
wonderflow, line 389:
```suggestion
// HistoryWorkload contain the old component revision that are still running
```

Issue #222, opened 2020-09-27
[BUG] dependency won't work as expected in updating


wonderflow:
# Problem

```
    kind: ApplicationConfiguration
    spec:
      components:
      - componentName: my-web
        dataInputs:
        - fieldPaths: 
            - spec.fieldA
          valueFrom:
            dataOutputName: output-data
        traits:
        - trait:
            kind: echo
            spec:
              data: abc
          dataOutputs:
            - name: output-data
              fieldPath: status.data
              conditions:
              - op: eq
                value: True
                fieldPath: status.Ready
```

# Background

OAM lacks of mechanism to indicate whether workload/trait CR is successfully observed by its controller.

So when an AppConfig CR is created, we can't confirm whether the update is resolved(by workload/trait controller) or not.

This is a little like K8s observedGeneration mechanism, and I used to raise [an issue to add observedGeneration for AppConfig and Component CR](https://github.com/crossplane/oam-kubernetes-runtime/issues/91).

While `observedGeneration` is not enough. Like the picture below, the `observedGeneration` can't ensure workload/trait reconciled successfully.

![image](https://user-images.githubusercontent.com/2173670/94360637-6c53b980-00e1-11eb-8155-39c3cbf85f9c.png)

# Proposal

calculate AppConfig.Spec hash and propagate as label into trait and workload

Diddaa:
Problem described greatly clear,  it just like CAS problem, we need Compare version before make any actions.

hongchaodeng:
I will fix the issue as soon as I can. Will pull up a PR in about two days.
The fix will be based on observed generation and will put a generation label into Workload/Trait CR.

For example, if an appConfig has:
```yaml
kind: AppConfig
metadata:
  generation: 1
```

Then its children CR will have:
```yaml
kind: ExampleTrait
metadata:
  labels:
    core.oam.dev/app-generation: "1"
```

Once the generation doesn't match the dependency engine will keep reconciling.

wonderflow:
I have sent a PR about this mechanism: #223

> Add valueFrom fieldPath for datainput/output condition, let oam-runtime check that value from metatdata.label should equal to status.observedApp

 ```
 dataOutputs:
   - name: output-data
     fieldPath: status.data
     conditions:
     - op: eq
       valueFrom:
          fieldPath: .metadata.labels.oam-app-hash
       fieldPath: status.observedApp
```

PR #223, opened 2020-09-28
add valueFrom AppConfig in depedency requirement

wonderflow:
This is one part to fix issue #222

1. add valueFrom in dependency requirement condition, that you can specify field path of AppConfig.
2. add reason message when requirement dependency not match


apis/core/v1alpha2/core_types.go
```diff
@@ -488,11 +488,22 @@ type DataInputValueFrom struct {
...
+	// Value is mutually exclusive with ValueFrom
+	Value string `json:"value,omitempty"`
+	// +optional
+	// This is mutually exclusive with Value
+	ValueFrom ValueFrom `json:"valueFrom,omitempty"`
 	// +optional
 	FieldPath string `json:"fieldPath,omitempty"`
 }
 
+// ValueFrom get value from AppConfig object by specify a path
```
hongchaodeng, line 502:
```suggestion
// ValueFrom gets value from AppConfig object by specifying a path
```

hongchaodeng (approved):
I would left LGTM here since it is mostly ready. Merge at your confidence.

PR #224, opened 2020-09-28
Use AppConfig's generation to get latest dependency only 

### LINKED ISSUE #222: [BUG] dependency won't work as expected in updating

hongchaodeng:
- pass AC's generation into child objects' labels
- check generation in getting data input

fix #222


pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -494,12 +496,23 @@ func (r *components) getDataInput(ctx context.Context, s *dagSource, ac *unstruc
...
+	var acGeneration int
+	if err := pavedAC.GetValueInto("metadata.generation", &acGeneration); err != nil && !fieldpath.IsNotFound(err) {
+		return nil, false, err.Error(), err
+	}
 
+	// The source object's app generation should match current AC. Otherwise it is from an old AC and we should wait until
+	// it is updated then reconcile again.
+	if g1, g2 := u.GetLabels()[oam.LabelAppGeneration], strconv.Itoa(acGeneration); g1 != g2 {
+		return nil, false, fmt.Sprintf("generation not match: obj (%s), ac (%s)", g1, g2), nil
+	}
```
wonderflow, line 513: The object can be workload or trait.

1. As a workload, the object will always have this label, it is meaningless to check.
2. As a trait, we have no way to guarantee the object has this label, in fact, no trait will have it.

And also, this is a break change, I don't think we need to make it as a force check.

pkg/controller/v1alpha2/applicationconfiguration/render_test.go
```diff
@@ -214,6 +215,7 @@ func TestRenderComponents(t *testing.T) {
 									oam.LabelAppName:              acName,
 									oam.LabelAppComponentRevision: "",
 									oam.LabelOAMResourceType:      oam.ResourceTypeTrait,
+									oam.LabelAppGeneration:        "0",
```
wonderflow, line 218: I don't see any code will add this label, why the test case will have this label if we don't manuelly add here?

wonderflow (changes requested):

hongchaodeng:
After discussion, this PR will not fix #222 because it also requires the trait operator to update the data in trait CR. So waiting for label to be updated is still not enough. Therefore, the data flow involves both OAM Runtime and the Trait Operators.

We will put this on hold and come back to this once we have an ideal solution.

Issue #218, opened 2020-09-24
[Feature] Run with fewer privileges


negz:
**Is your feature request related to a problem? Please describe.**

https://github.com/crossplane/oam-kubernetes-runtime/blob/master/charts/oam-kubernetes-runtime/templates/oam-controller.yaml#L19

Currently the runtime runs as `cluster-admin` - i.e. it has unlimited access to the cluster. I presume this is because the AppConfig controller needs to create workloads and scopes, which could be of arbitrary kinds (depending on what workload and scope definitions exist).

**Describe the solution you'd like**

I believe a freshly installed OAM runtime supports only the core workloads and traits. Would it be possible to bind to a more narrowly scoped cluster role that had only permissions to CRUD core workloads and traits? Infra operators could update this cluster role if and when they wanted to allow the runtime to work with new workload or trait types.

**Describe alternatives you've considered**

Other options include:

* Allow the workload and scope types that should be supported to be configurable by the Helm chart - i.e. the chart would allow creating extended workload and trait definitions, and would also ensure the controller's cluster role allowed it access to CRUD the relevant types.
* Introduce something like the [Crossplane RBAC manager](https://github.com/crossplane/crossplane/blob/master/design/design-doc-rbac-manager.md). This process would have `bind` and `escalate` RBAC access (it can grant access it does not have), and would automatically grant the OAM runtime access to reconcile types that were referenced by a workload or trait definition.

PR #228, opened 2020-10-06
Run with fewer privileges

negz:
Fixes https://github.com/crossplane/oam-kubernetes-runtime/issues/218

This commit updates the Helm chart to avoid running as cluster-admin. Instead, the controller runs only with the privileges it needs 'out of the box'; i.e. to manage all core OAM types, as well as deployments and services.

The commit also includes a few small chart hygiene fixes; i.e. ensuring that names will not collide when multiple releases exist in the same cluster, and that all resources include the standard labels.

### DIFF
--- a/charts/oam-kubernetes-runtime/templates/oam-controller.yaml
+++ b/charts/oam-kubernetes-runtime/templates/oam-controller.yaml
@@ -5,18 +5,60 @@ kind: ServiceAccount
 metadata:
   name: {{ include "oam-kubernetes-runtime.serviceAccountName" . }}
   labels:
-  {{ include "oam-kubernetes-runtime.labels" . | nindent 4 }}
-  {{- end }}
----
+  {{- include "oam-kubernetes-runtime.labels" . | nindent 4 }}
+{{- end }}
 
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "oam-kubernetes-runtime.fullname" . }}
+  labels:
+  {{- include "oam-kubernetes-runtime.labels" . | nindent 4 }}
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.oam.dev/aggregate-to-controller: "true"
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "oam-kubernetes-runtime.fullname" . }}:system:aggregate-to-controller
+  labels:
+    {{- include "oam-kubernetes-runtime.labels" . | nindent 4 }}
+    rbac.oam.dev/aggregate-to-controller: "true"
+rules:
+- apiGroups:
+  - core.oam.dev
+  resources:
+  - "*"
+  verbs:
+  - "*"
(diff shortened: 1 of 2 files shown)


charts/oam-kubernetes-runtime/templates/oam-controller.yaml
```diff
@@ -4,19 +4,58 @@ apiVersion: v1
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "oam-kubernetes-runtime.fullname" . }}
+  labels: {{ include "oam-kubernetes-runtime.labels" . | nindent 4 }}
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.oam.dev/aggregate-to-controller: "true"
```
negz, line 19: I elected to use cluster role aggregation here so that it would be easy for users to extend the privileges of the OAM controller. I was thinking this could help with supporting new non-core kinds of workloads and traits; e.g. if those kinds were created by the core AppConfig controller, but reconciled by controllers running as distinct deployments in the cluster.

negz:
I believe this is good to go - I've tested it by updating the e2e tests to run with only the privileges the Helm chart grants them (plus an additional aggregate role that grants access to `example.com` resources).

hasheddan (approved):

wonderflow:
Thanks @negz ! This PR is good, but I wonder it would be hard for user to register a new CRD to work as an OAM trait/workload.

Currently, only a WorkloadDefinition/TraitDefinition is needed. Can you give an example workflow if we add this `aggregate ` mechanism to register a new CRD, it would also help us to understand this PR better!

negz:
> I wonder it would be hard for user to register a new CRD to work as an OAM trait/workload.

> Can you give an example workflow if we add this aggregate mechanism to register a new CRD, it would also help us to understand this PR better!

The process is identical, except that the person who authors the CRD and the WorkloadDefinition (or ScopeDefinition) must also author a ClusterRole that grants the OAM controller access to that type, so that the AppConfig controller can create, update, and delete it as necessary. This seems like a critical best practice to me; we should not run our controllers as cluster-admin.

The example in the e2e tests may be a good illustration of this. We create the CRD and WorkloadDefinition for `foo.example.com` so that it may be used as a workload. At this point the OAM controller does not have RBAC access to use a `kind: Foo`, so we must also grant that access via a `ClusterRole`:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: foo.example.com
    rbac.oam.dev/aggregate-to-controller: "true"
rules:
- apiGroups:
  - "example.com"
  resources:
  - foo
  verbs:
  - "*"
```

https://crossplane.slack.com/archives/CTHADJCEN/p1600977951000500

This approach was discussed in the above Slack conversation. There may be further automation we can do in future, like Crossplane's RBAC manager, but for now I feel it's a good compromise to allow us to feel comfortable including oam-kubernetes-runtime with Crossplane v0.13 later this week.

ryanzhang-oss (approved):
Thank you for working on this

wonderflow (approved):
Thanks @negz

PR #232, opened 2020-10-08
Update apply.go

xishengcai:
如果自定义的scope 对象 第一次创建的时候， 没有加入spec字段， 会报错，此时不应该return， 而应该继续向下执行， 第一次patch 成功后， 这一段 pave 就不会报错，然后检查workload ref 是否重复。

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/apply.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/apply.go
@@ -222,8 +222,10 @@ func (a *workloads) applyScope(ctx context.Context, wl Workload, s unstructured.
 				return nil
 			}
 		}
-	} else {
-		return errors.Wrapf(err, errFmtGetScopeWorkloadRef, s.GetAPIVersion(), s.GetKind(), s.GetName(), workloadRefsPath)
+	}else if err.Error() == "spec: no such field"{
+            // init scope, so not found field: spec
+	}else {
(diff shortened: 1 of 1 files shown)


pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -222,8 +222,10 @@ func (a *workloads) applyScope(ctx context.Context, wl Workload, s unstructured.
 				return nil
 			}
 		}
-	} else {
-		return errors.Wrapf(err, errFmtGetScopeWorkloadRef, s.GetAPIVersion(), s.GetKind(), s.GetName(), workloadRefsPath)
+	}else if err.Error() == "spec: no such field"{
```
ryanzhang-oss, line 225: compare err with some magic string is an anti-pattern.

ryanzhang-oss: We can't proceed if the workloadRef path is provided but doesn't work

ryanzhang-oss (changes requested):
We would like to understand the behavior better with a full explanation in English and clean code with tests.

captainroy-hy:
AppConfig depends on the field path from ScopeDefinition's `spec.workloadRefs` to pass workload references to scope controllers. So if that field is missed, AppConfig controller is supposed to return error. It's not responsible for AppConfig controller to handle this through patching HealthScope forcefully.

PR #236, opened 2020-10-09
RBAC rules must use plural resource names

negz:
I was end-to-end testing our Crossplane documentation prior to the release tomorrow, and noticed that the OAM pod was complaining that it didn't have access to deployments. When I change these to plurals, the pods stop complaining.


negz:
I believe this might require an 0.2.1 release of OAM, which we'll need to pin at https://github.com/crossplane/crossplane/pull/1818.

hasheddan (approved):

PR #240, opened 2020-10-09
add podSpecPath into workloadDefinition

### LINKED ISSUE #233: [Proposal] Add a 'podSpecPath' field in WorkloadDefinition
We have discussed a lot and finally conclude to add labels like [`workload.oam.dev/podspecable`](https://github.com/oam-dev/spec/blob/master/4.workload_definitions.md#labels) to indicate the workload will contain 'podSpec' in it's spec.

But it's hard to know which field is podSpec, so I propose we add a 'podSpecPath' field in WorkloadDefinition.

By defualt, if `workload.oam.dev/podspecable: true` label is set but `podSpecPath` not, we may go to `spec.template` to find 'podTemplate', and get 'podSpec' in `spec.template.spec`.

If this field is set like below, we will directly regard path specified in `podSpecPath` as `podSpec` path, in this example, it's `spec.podSpec`

```
kind: WorkloadDefinition
metadata:
  name: webservice
  labels:
     workload.oam.dev/podspecable: true
  annotations:
    definition.oam.dev/apiVersion: "standard.oam.dev/v1alpha1"
    definition.oam.dev/kind: "PodSpecWorkload"
spec:
  podSpecPath: "spec.podSpec"
  definitionRef:
    name: podspecworkloads.standard.oam.dev
  childResourceKinds:
    - apiVersion: apps/v1
      kind: Deployment
    - apiVersion: v1
      kind: Service
```

This is also needed to fix https://github.com/oam-dev/kubevela/issues/354

wonderflow:
fixes #233

Please read that issue for more detail.

### DIFF
--- a/apis/core/v1alpha2/core_types.go
+++ b/apis/core/v1alpha2/core_types.go
@@ -54,6 +54,11 @@ type WorkloadDefinitionSpec struct {
 	// +optional
 	RevisionLabel string `json:"revisionLabel,omitempty"`
 
+	// PodSpecPath indicates where/if this workload has K8s podSpec field
+	// if one workload has podSpec, trait can do lot's of assumption such as port, env, volume fields.
+	// +optional
+	PodSpecPath string `json:"podSpecPath,omitempty"`
+
 	// Extension is used for extension needs by OAM platform builders
 	// +optional
 	// +kubebuilder:pruning:PreserveUnknownFields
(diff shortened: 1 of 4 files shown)


ryanzhang-oss (commented):
Why combine the design and the api only part of the implementation together?

ryanzhang-oss (approved):

PR #242, opened 2020-10-09
add mutating for appConfig to support name/properties format of trait

### LINKED ISSUE #213: [BUG] provide a way to help find a TraitDefintion whose name doesn't follow the convention of <plurals-kind>.<resource-group>
Similar to https://github.com/crossplane/oam-kubernetes-runtime/issues/207, we need to provide a way to help find a TraitDefintion whose name doesn't follow the convention of `<plurals-kind>.<resource-group>`

wonderflow:
1. align with the [new trait format design of spec](https://github.com/oam-dev/spec/blob/master/7.application_configuration.md#trait)
2. this PR will also fixes #213 a label added to find traitDefinition easy.

### DIFF
--- a/pkg/oam/labels.go
+++ b/pkg/oam/labels.go
@@ -27,6 +27,11 @@ const (
 	LabelAppComponentRevision = "app.oam.dev/revision"
 	// LabelOAMResourceType whether a CR is workload or trait
 	LabelOAMResourceType = "app.oam.dev/resourceType"
+
+	// WorkloadTypeLabel indicates the type of the workloadDefinition
+	WorkloadTypeLabel = "workload.oam.dev/type"
+	// TraitTypeLabel indicates the type of the traitDefinition
+	TraitTypeLabel = "trait.oam.dev/type"
 )
 
 const (
--- a/pkg/oam/util/helper_test.go
+++ b/pkg/oam/util/helper_test.go
@@ -698,8 +698,9 @@ var _ = Describe("Test workload related helper utils", func() {
 var _ = Describe("Test unstructured related helper utils", func() {
 	It("Test get CRD name from an unstructured object", func() {
 		tests := map[string]struct {
-			u   *unstructured.Unstructured
-			exp string
+			u         *unstructured.Unstructured
+			typeLabel string
+			exp       string
 		}{
 			"native resource": {
 				u: &unstructured.Unstructured{Object: map[string]interface{}{
@@ -715,11 +716,37 @@ var _ = Describe("Test unstructured related helper utils", func() {
 				}},
 				exp: "simplerollouttraits.extend.oam.dev",
 			},
+			"trait": {
+				u: &unstructured.Unstructured{Object: map[string]interface{}{
+					"apiVersion": "extend.oam.dev/v1alpha2",
+					"kind":       "SimpleRolloutTrait",
+					"metadata": map[string]interface{}{
+						"labels": map[string]interface{}{
+							oam.TraitTypeLabel: "rollout",
+						},
+					},
+				}},
+				typeLabel: oam.TraitTypeLabel,
+				exp:       "rollout",
+			},
+			"workload": {
+				u: &unstructured.Unstructured{Object: map[string]interface{}{
+					"apiVersion": "apps/v1",
+					"kind":       "Deployment",
+					"metadata": map[string]interface{}{
+						"labels": map[string]interface{}{
+							oam.WorkloadTypeLabel: "deploy",
+						},
+					},
+				}},
+				typeLabel: oam.WorkloadTypeLabel,
+				exp:       "deploy",
+			},
 		}
 		for name, ti := range tests {
-			got := util.GetDefinitionName(ti.u)
+			got := util.GetDefinitionName(ti.u, ti.typeLabel)
 			By(fmt.Sprint("Running test: ", name))
-			Expect(ti.exp).Should(Equal(got))
+			Expect(got).Should(Equal(ti.exp))
 		}
 	})
 })
--- a/pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
+++ b/pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
@@ -1,0 +1,174 @@
+package applicationconfiguration
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"reflect"
+
+	"github.com/davecgh/go-spew/spew"
+	crdv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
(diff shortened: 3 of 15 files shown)


hongchaodeng:
> new trait format design of spec

This link is not working. Where is the design?

wonderflow:
@hongchaodeng  Sorry for the borken link, fixed now. It's https://github.com/oam-dev/spec/blob/master/7.application_configuration.md#trait

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
```diff
@@ -0,0 +1,174 @@
...
+	if content[TraitSpecField] != nil {
+		content["spec"] = content[TraitSpecField]
+		delete(content, TraitSpecField)
+	}
+
+	trait := unstructured.Unstructured{
+		Object: content,
+	}
+	// find out the GVK from the CRD definition and set
+	apiVersion := metav1.GroupVersion{
```
resouer, line 136: I am OK for short term fix by simply replacing name/properties with GVK. In long term it's better to fix this in AppConfig controller so `$ kubectl get appconfig` will also return name/properties format.

pkg/oam/util/helper_test.go
```diff
@@ -715,11 +716,37 @@ var _ = Describe("Test unstructured related helper utils", func() {
 				}},
 				exp: "simplerollouttraits.extend.oam.dev",
 			},
+			"trait": {
+				u: &unstructured.Unstructured{Object: map[string]interface{}{
+					"apiVersion": "extend.oam.dev/v1alpha2",
+					"kind":       "SimpleRolloutTrait",
+					"metadata": map[string]interface{}{
```
ryanzhang-oss, line 723: it will be good to add a name here that is different from the label so we know which is the truth

pkg/oam/util/helper_test.go
```diff
@@ -715,11 +716,37 @@ var _ = Describe("Test unstructured related helper utils", func() {
...
+				}},
+				typeLabel: oam.TraitTypeLabel,
+				exp:       "rollout",
+			},
+			"workload": {
+				u: &unstructured.Unstructured{Object: map[string]interface{}{
+					"apiVersion": "apps/v1",
+					"kind":       "Deployment",
+					"metadata": map[string]interface{}{
+						"labels": map[string]interface{}{
```
ryanzhang-oss, line 737: it will be good to add a name here that is different from the label so we know which is the truth

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return resp
+}
+
+// Mutate sets all the default value for the Component
+func (h *MutatingHandler) Mutate(obj *v1alpha2.ApplicationConfiguration) error {
+	mutatelog.Info("mutate", "name", obj.Name)
+
+	for compIdx, comp := range obj.Spec.Components {
+		var updated bool
```
ryanzhang-oss, line 80: do we really need this?  Can't we just blindly copy back no matter what? The logic will be much simplified

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
ryanzhang-oss: Yes, this case is different from the `component` mutate hook as that was completely inside the component. I preferred to handle the explicit field in the controller instead. It's actually simpler to test.

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
```diff
@@ -0,0 +1,174 @@
...
+	// find out the GVK from the CRD definition and set
+	apiVersion := metav1.GroupVersion{
+		Group:   customResourceDefinition.Spec.Group,
+		Version: customResourceDefinition.Spec.Versions[0].Name,
+	}.String()
+	trait.SetAPIVersion(apiVersion)
+	trait.SetKind(customResourceDefinition.Spec.Names.Kind)
+	mutatelog.Info("Set the trait GVK", "trait api version", trait.GetAPIVersion(), "trait Kind", trait.GetKind())
+	// copy namespace/label/annotation to the trait and add traitType label
+	trait.SetNamespace(namespace)
```
ryanzhang-oss, line 144: why is the namespace needed to be copied? Is it not in the content before?

pkg/oam/labels.go
```diff
@@ -27,6 +27,11 @@ const (
 	LabelAppComponentRevision = "app.oam.dev/revision"
 	// LabelOAMResourceType whether a CR is workload or trait
 	LabelOAMResourceType = "app.oam.dev/resourceType"
+
+	// WorkloadTypeLabel indicates the type of the workloadDefinition
+	WorkloadTypeLabel = "workload.oam.dev/type"
+	// TraitTypeLabel indicates the type of the traitDefinition
+	TraitTypeLabel = "trait.oam.dev/type"
```
captainroy-hy, line 34: A little bit inconsistent: In AppConfig, the field name is `spec.components[x].traits[y].name` but the lable key is `type`.

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -65,18 +68,56 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
 		}
+		// TODO(wonderflow): Add more validation logic here.
 	}
 	return admission.ValidationResponse(true, "")
 }
 
+// ValidateTraitObject validates the ApplicationConfiguration on creation/update
+func ValidateTraitObject(obj *v1alpha2.ApplicationConfiguration) field.ErrorList {
+	klog.Info("validate applicationConfiguration", "name", obj.Name)
+	allErrs := apimachineryvalidation.ValidateObjectMeta(&obj.ObjectMeta, true,
```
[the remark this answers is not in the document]
wonderflow: Leave it here is OK, this is a validator for the whole appConfig, but some parts are missing, we can add them, in the future.

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -65,18 +68,56 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
+		apimachineryvalidation.NameIsDNSSubdomain, field.NewPath("metadata"))
+	for cidx, comp := range obj.Spec.Components {
+		for idx, tr := range comp.Traits {
+			fldPath := field.NewPath("spec").Child("components").Index(cidx).Child("traits").Index(idx).Child("trait")
```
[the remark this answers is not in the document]
wonderflow: This structure is `OAM spec`， no one can easily change it. If changed, there should be a new webhook version.

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
wonderflow: I don't think add this flag makes any complex here. I don't feel bindly copy back is good.

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
wonderflow: yes， I shouldn't add it here, thanks, will fix it.

pkg/webhook/v1alpha2/applicationconfiguration/mutating_handler.go
wonderflow: `component` mutate hook and this PR are actually the same, I hope we can use a separate PR to move them inside the controller together **if we really want to do that**.

pkg/oam/util/helper_test.go
wonderflow: I don't know what do you mean here? Which name?

pkg/oam/labels.go
wonderflow: Yes, but it's actually a type, I'm aligning with workload here.

wonderflow:
@ryanzhang-oss all comments are replied/fixed, please review again, thanks

ryanzhang-oss:
Synced off-line with @wonderflow,  we realized that the mutation is not really addressing the new spec that uses the trait name to point to the `TraitDefinition`. However, we decided that it's still useful to have an applicationConfig mutation framework as we will surely need one later.

ryanzhang-oss (approved):

PR #245, opened 2020-10-10
Fix for loop

LeoLiuYan:
Signed-off-by: LeoLiuYan <929908264@qq.com>

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render.go
@@ -445,7 +445,7 @@ func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInp
 		}
 		if !ready {
 			uds = append(uds, makeUnsatisfiedDependency(obj, s, in, reason))
-			return uds, nil
+			continue
 		}
 
 		err = fillValue(obj, in.ToFieldPaths, val)
@@ -619,8 +619,10 @@ func getTraitName(ac *v1alpha2.ApplicationConfiguration, componentName string,
 		for _, trait := range w.Traits {
 			if trait.Reference.APIVersion == apiVersion && trait.Reference.Kind == kind {
 				traitName = trait.Reference.Name
+				break
 			}
 		}
+		break
 	}
 
 	if len(traitName) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -445,7 +445,7 @@ func (r *components) handleDataInput(ctx context.Context, ins []v1alpha2.DataInp
 		}
 		if !ready {
 			uds = append(uds, makeUnsatisfiedDependency(obj, s, in, reason))
-			return uds, nil
+			continue
```
wonderflow, line 448: It makes sense when dependency condition unsatisfied, can you give a case to show why we MUST use `continue` here? Thanks.

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -619,8 +619,10 @@ func getTraitName(ac *v1alpha2.ApplicationConfiguration, componentName string,
 		for _, trait := range w.Traits {
 			if trait.Reference.APIVersion == apiVersion && trait.Reference.Kind == kind {
 				traitName = trait.Reference.Name
```
wonderflow, line 621: just return this value is better than break.

wonderflow (changes requested):

pkg/controller/v1alpha2/applicationconfiguration/render.go
Ghostbaby: it didn't need to loop check other `v1alpha2.DataInput` ？

wonderflow: this return with condition unsatisfied will lead to another check/reconcile soon.

LeoLiuYan: ok

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -619,6 +619,7 @@ func getTraitName(ac *v1alpha2.ApplicationConfiguration, componentName string,
...
+				return traitName
```
wonderflow, line 622: I think this PR is not necessary, and it will cause bug if we return here. because traitName here could be empty.


What's the most need here is a test case. 😄

LeoLiuYan: So add 'break' here is a good way.

Issue #202, opened 2020-09-08
[Feature] Using K8s Discovery mechanism to get resource from GVK


wonderflow:
```
// GetCRDName return the CRD name of any resources
// the format of the CRD of a resource is <kind purals>.<group>
func GetCRDName(u *unstructured.Unstructured) string {
	group, _ := APIVersion2GroupVersion(u.GetAPIVersion())
	resources := []string{Kind2Resource(u.GetKind())}
	if group != "" {
		resources = append(resources, group)
	}
	return strings.Join(resources, ".")
}
// Kind2Resource convert Kind to Resources
func Kind2Resource(str string) string {
	return plur.NewClient().Plural(strings.ToLower(str))
}
```

Now we convert GVK to CRD Name by using plural of kind. 

We should get the resource by [using discovery client](https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L187-L209).

The `ServerResourcesForGroupVersion` can get resourceList from groupVersion(apiVersion) and we can get the real resource from the list.

PR #246, opened 2020-10-10
fix miss rbac about controllerrevision and CRD

wonderflow:
- CRD is needed when OAM is doing mutating
- controllerrevision is needed for versioning mechanisim.

@negz @hasheddan

### DIFF
--- a/charts/oam-kubernetes-runtime/templates/oam-controller.yaml
+++ b/charts/oam-kubernetes-runtime/templates/oam-controller.yaml
@@ -39,6 +39,13 @@ rules:
   - apps
   resources:
   - deployments
+  - controllerrevisions
+  verbs:
+  - "*"
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
   verbs:
   - "*"
 - apiGroups:


charts/oam-kubernetes-runtime/templates/oam-controller.yaml
```diff
@@ -39,6 +39,13 @@ rules:
   - apps
   resources:
   - deployments
+  - controllerrevisions
+  verbs:
+  - "*"
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
```
hasheddan, line 48: Is this still needed when webhook is not deployed? Should we make it conditional on that?

wonderflow: yes, i think we will still need it for Definition -> CRD Reference logic

ryanzhang-oss (approved):

charts/oam-kubernetes-runtime/templates/oam-controller.yaml
negz: > i think we will still need it for Definition -> CRD Reference logic even the webhook not open.

Are you sure? I can't see any evidence of that: https://github.com/crossplane/oam-kubernetes-runtime/search?l=Go&q=CustomResourceDefinition

wonderflow: I'm thinking of this issue https://github.com/crossplane/oam-kubernetes-runtime/issues/202 @negz

PR #249, opened 2020-10-15
minor tweak, assign the workloadRef during render instead of apply

ryanzhang-oss:
Assign the workloadRef during render instead of apply phase. The result is a more clean-cut division between apply and render and one search for the traitDefinition instead of two.

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render.go
@@ -179,7 +179,6 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 		trait := traits[i]
 		workloadRefPath := traitDef.Spec.WorkloadRefPath
 		if len(workloadRefPath) != 0 {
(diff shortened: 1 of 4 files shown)


wonderflow (approved):

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -163,7 +167,24 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
...
+		Kind:       w.GetKind(),
+		Name:       w.GetName(),
+	}
+	//  We only patch a TypedReference object to the trait if it asks for it
+	for i := range acc.Traits {
+		traitDef := traitDefs[i]
+		trait := traits[i]
+		workloadRefPath := traitDef.Spec.WorkloadRefPath
+		if len(workloadRefPath) != 0 {
+			fmt.Printf("\n\n\nworkloadRef path = %s\n\n\n", workloadRefPath)
```
wonderflow, line 182: please remove this debug info

ryanzhang-oss: done, good catch

Issue #251, opened 2020-10-16
[Feature] Optimization of cleanupControllerRevision


linjiemiao:
There is a problem with the cleanupControllerRevision call logic in createControllerRevision. Currently:
if int64(c.RevisionLimit) <nextRevision {
     1) Get the ControllerRevision list corresponding to the component
     2) Get the applicationConfiguration list
     3) Calculate the number of gc ControllerRevision required, revisionName mapping  and sort the revisions
     4) Delete redundant revision
}

**Describe the solution you'd like**

After 1)  is completed, judge len(revisions) <RevisionLimit; return directly.

linjiemiao:
If the nextRevision of controllerrevision exceeds RevisionLimit, then the number of controllerrevisions currently in the system is : len(revisionNum) + len(liveHashes), this value must be greater than 50, so no additional judgment is required.

linjiemiao:
I found a bug, the sortedControllerRevision function has some problems when calculating liveHashes. toKeep = revisionLimit + len(this component use RevisionName). So we need to pass the name of the current component to sortedControllerRevision. When traversing AppConfig, we need to judge ExtractComponentName(component.RevisionName) == currentComponentName, and then we can set its liveHashes.

PR #252, opened 2020-10-16
[ISSUE #251] [Feature] Optimization of cleanupControllerRevision

linjiemiao:
I found a bug, the sortedControllerRevision function has some problems when calculating liveHashes. toKeep = revisionLimit + len(this component use RevisionName). So we need to pass the name of the current component to sortedControllerRevision. When traversing AppConfig, we need to judge ExtractComponentName(component.RevisionName) == currentComponentName, and then we can set its liveHashes.

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/component.go
@@ -193,7 +193,7 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 
 // get sorted controllerRevisions, prepare to delete controllerRevisions
 func sortedControllerRevision(appConfigs []v1alpha2.ApplicationConfiguration, revisions []appsv1.ControllerRevision,
-	revisionLimit int) (sortedRevisions []appsv1.ControllerRevision, toKill int, liveHashes map[string]bool) {
+	revisionLimit int, currentComponentName string) (sortedRevisions []appsv1.ControllerRevision, toKill int, liveHashes map[string]bool) {
 	liveHashes = make(map[string]bool)
 	sortedRevisions = revisions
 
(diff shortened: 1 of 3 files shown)


captainroy-hy:
You mean, currently, liveHashes records revision names used by ALL components of ALL AppConfig, but it's supposed to record revision name ONLY used by the current component, right?

linjiemiao:
> You mean, currently, liveHashes records revision names used by ALL components of ALL AppConfig, but it's supposed to record revision name ONLY used by the current component, right?

Yes， this is my propose

ryanzhang-oss:
Please add tests that repro this bug

linjiemiao:
> Please add tests that repro this bug

ok, I will add tests later

linjiemiao:
Why is it that when I execute make e2e test, it randomly fails to pass all of them, and all the errors are prompting timeout. Is it because the load of the test environment for executing e2e is too high?

wonderflow:
@linjiemiao yes, sometimes the CI fails of timeout, maybe because of the test environment lack of resources, please re-run it.

linjiemiao:
I have update test cases, please review my code。Still Go/Build(pull_request)[make e2e test] always timeout.

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -193,15 +193,17 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
...
+	revisionLimit int, currentComponentName string) (sortedRevisions []appsv1.ControllerRevision, toKill int, liveHashes map[string]bool) {
 	liveHashes = make(map[string]bool)
 	sortedRevisions = revisions
 
 	// get all revisions used and skipped
 	for _, appConfig := range appConfigs {
 		for _, component := range appConfig.Spec.Components {
 			if component.RevisionName != "" {
-				liveHashes[component.RevisionName] = true
+				if ExtractComponentName(component.RevisionName) == currentComponentName {
```
wonderflow, line 204: I don't think this is necessary, because we use label selector to filter the revisions, so the revisions passed into this function are all related to the specific Component only. https://github.com/crossplane/oam-kubernetes-runtime/blob/master/pkg/controller/v1alpha2/applicationconfiguration/component.go#L226


So there maybe no bug, right?

linjiemiao: yes, no bug. i close this issue.

PR #257, opened 2020-10-19
minor fix datarace

wonderflow:


hongchaodeng (approved):
lgtm

pkg/controller/v1alpha2/applicationconfiguration/component.go
```diff
@@ -133,38 +133,39 @@ func newTrue() *bool {
 
 func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtime.Object) bool {
 	curComp := obj.(*v1alpha2.Component)
-	diff, curRevision := c.IsRevisionDiff(mt, curComp)
+	comp := curComp.DeepCopy()
```
ryanzhang-oss, line 136: It seems the controller runtime already returns a copy from the informer queue(client go, tools/cache/delta_fifo.go:486), is it?

wonderflow: I'm afraid not, currently vela complains the error: https://github.com/oam-dev/kubevela/runs/1272425015

After debug, I found out it because of this not copy.

hongchaodeng: Well, the event handler passes the object reference directly:

```go
// OnAdd creates CreateEvent and calls Create on EventHandler
func (e EventHandler) OnAdd(obj interface{}) {
	...
	// Pull the runtime.Object out of the object
	if o, ok := obj.(runtime.Object); ok {
		c.Object = o
	} 

```

PR #260, opened 2020-10-21
Migrate from Jenkins to GitHub Actions

negz:
See https://github.com/crossplane/crossplane/pull/1865 for context. Note that I've also attempted to merge the existing GitHub e2e workflow into the standard one we're using for all Crossplane projects, since the existing one was doing duplicate work.

If these changes look good, I will:

* Remove the webhook that triggers the Jenkins build
* Mark the Jenkins and 'Build' checks as not required
* Mark the lint, check-diff, and unit-tests checks as required

I notice the e2e tests seem flaky. Should I also mark them required?


.github/workflows/ci.yml
```diff
@@ -0,0 +1,252 @@
...
+      - name: Vendor Dependencies
+        run: make vendor vendor.check
+
+      - name: Build Helm Chart
+        run: make -j2 build
+        env:
+          # We're using docker buildx, which doesn't actually load the images it
+          # builds by default. Specifying --load does so.
+          BUILD_ARGS: "--load"
+
```
hasheddan, line 169: Looks like just adding https://github.com/crossplane/oam-kubernetes-runtime/blob/723f8ecf8444e0feed1493156626df6d54d0c116/.github/workflows/go.yml#L46 may do the trick

wonderflow:
This PR is really cool! @negz Thanks!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,279 @@
...
+      - name: Vendor Dependencies
+        run: make vendor vendor.check
+
+      - name: Check Diff
+        run: make check-diff
+
+  unit-tests:
+    runs-on: ubuntu-18.04
+    needs: detect-noop
+    if: needs.detect-noop.outputs.noop != 'true'
```
wonderflow, line 107: can you explain a bit what will it help? I see lots of this lines, but not very understand why we need it.  thanks

wonderflow (approved):
Thanks @negz , this is really cool Job. 

e2e test is also required, we don't know why it's sometimes broken, we can't re-produce it locally, I assume it's because of lack of resource?

.github/workflows/ci.yml
negz: This is just a little optimisation that skips running CI (i.e. automatically passes CI) for PRs that either:

* Only touch markdown files or images.
* Are identical to another PR that already passed CI (i.e. the content of the two refs would have matching hashes).

PR #261, opened 2020-10-22
get CRD by discovery mechanism instead of get plural

### LINKED ISSUE #202: [Feature] Using K8s Discovery mechanism to get resource from GVK

wonderflow:
1. use discovery mechanism to discovery definition name by using CRD name  fix #202
2. refactor the test, it's mass and I fixed them by refactoring.


ryanzhang-oss:
Why is the build module changed? Is that even supposed to be checked-in?

pkg/oam/util/helper.go
```diff
@@ -230,35 +238,23 @@ func PassLabelAndAnnotation(parentObj oam.Object, childObj labelAnnotationObject
 // the format of the definition of a resource is <kind plurals>.<group>
 // Now the definition name of a resource could also be defined as `definition.oam.dev/name` in `metadata.annotations`
 // typeLabel specified which Definition it is, if specified, will directly get definition from label.
-func GetDefinitionName(u *unstructured.Unstructured, typeLabel string) string {
+func GetDefinitionName(dm discoverymapper.DiscoveryMapper, u *unstructured.Unstructured, typeLabel string) (string, error) {
```
ryanzhang-oss, line 241: I know it's kinda late to change but it seems that it is better to make the dm a public variable in the package inside the init() function? In this way, we don't need to pass this around, all the caller of the utility functions shouldn't need to create a DiscoveryMapper. This is a pubic utility function that is actually used by other repros, this can break cross repo dependency.

test/e2e-test/suite_test.go
```diff
@@ -197,52 +197,6 @@ var _ = BeforeSuite(func(done Done) {
 	}
 	Expect(k8sClient.Create(context.Background(), &adminRoleBinding)).Should(BeNil())
 	By("Created cluster role binding for the test service account")
-	// Create a crd for appconfig dependency test
```
ryanzhang-oss, line 200: where did this part go?

wonderflow: it's moved out to dependency

ryanzhang-oss (approved):

PR #267, opened 2020-10-25
don't block app from running if no definition found

### LINKED ISSUE #82: Don't block if traitDefinition or workloadDefinition not exist
Now oam-k8s-runtime will check traitDefinition(for revisionAware) andworkloadDefinition(for workloadRef), but if these definition file not exist, application deploy will fail. However, the trait or workload can work without these information, so we shouldn't require definition files MUST exist. 

In fact, we can automatically generate a basic definition file for those workload/trait, or we can regard there's one. Then the deploy won't fail.

For example, for ingress trait as below:

```
...
      traits:
        - trait:
            apiVersion: extensions/v1beta1
            kind: Ingress
            metadata:
              name: tracker-ingress
            spec:
                backend:
                  serviceName: web-ui
                  servicePort: 8080 
```

If we don't have a traitdefinition, we can assume it's a basic one like below:

```
kind: TraitDefinition
metadata:
  name: ingresses.extensions
spec:
  definitionRef:
    name: ingresses.extensions 
```

All the information can be automatically generated. The main idea is to not block a normal deploy.

wonderflow:
fix  #82

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
@@ -42,6 +42,7 @@ import (
 	"github.com/crossplane/oam-kubernetes-runtime/pkg/controller"
 	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam"
 	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/discoverymapper"
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/util"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -453,7 +457,10 @@ func (w Workload) Status() v1alpha2.WorkloadStatus {
 		Traits: make([]v1alpha2.WorkloadTrait, len(w.Traits)),
 		Scopes: make([]v1alpha2.WorkloadScope, len(w.Scopes)),
 	}
-	for i := range w.Traits {
+	for i, tr := range w.Traits {
+		if tr.Definition.Name == util.Dummy && tr.Definition.Spec.Reference.Name == util.Dummy {
```
hongchaodeng, line 461: It seems that this only applies to traits. What about workloads?

wonderflow: we don't check workloadDef when apply worklaod.

hongchaodeng (approved):
lgtm

PR #268, opened 2020-10-25
add version into definitionReference for workload/trait/scope definitions

### LINKED ISSUE #210: [Feature] add apiversion and kind into Definition
In vela we write definition like this:

```
kind: TraitDefinition
metadata:
  name: metricstraits.standard.oam.dev
  namespace: default
  annotations:
    definition.oam.dev/apiVersion: standard.oam.dev/v1alpha1
    definition.oam.dev/kind: MetricsTrait
spec:
  appliesToWorkloads:
    - containerizedworkloads.core.oam.dev
    - clonesetworkloads.apps.kruise.io
  definitionRef:
    name: metricstraits.standard.oam.dev
  workloadRefPath: spec.workloadRef
```

We need to use annotations to tell which GVK the definition is pointing to, this is designed to be fixed by field `definitionRef`.

The original design is that we assume there will always be only one version for one existing CRD, but we can't actually.

Currently we use the first version found: https://github.com/crossplane/oam-kubernetes-runtime/blob/master/pkg/webhook/v1alpha2/component/mutating_handler.go#L120

This may not correct.

So we prepare to add below:

```
kind: TraitDefinition
metadata:
  name: metricstraits.standard.oam.dev
  namespace: default
spec:
  appliesToWorkloads:
    - containerizedworkloads.core.oam.dev
    - clonesetworkloads.apps.kruise.io
  definitionRef:
    name: metricstraits.standard.oam.dev
    apiVersion: standard.oam.dev/v1alpha1
    kind: MetricsTrait
  workloadRefPath: spec.workloadRef
```

This also works for WorkloadDefinition.

# Alternatives

If we only add a version into DefinitionRef, this also works, we can [discovery resources by group and version using discovery client](https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L90) 

```
kind: TraitDefinition
metadata:
  name: metricstraits.standard.oam.dev
  namespace: default
spec:
  appliesToWorkloads:
    - containerizedworkloads.core.oam.dev
    - clonesetworkloads.apps.kruise.io
  definitionRef:
    name: metricstraits.standard.oam.dev
    version: v1alpha1
  workloadRefPath: spec.workloadRef
```

wonderflow:
fix #210

### DIFF
--- a/pkg/oam/util/helper.go
+++ b/pkg/oam/util/helper.go
@@ -9,6 +9,8 @@ import (
 	"reflect"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
+
 	cpv1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/go-logr/logr"
@@ -295,6 +297,23 @@ func GetDefinitionName(dm discoverymapper.DiscoveryMapper, u *unstructured.Unstr
 		return "", err
 	}
 	return mapping.Resource.Resource + "." + groupVersion.Group, nil
+}
+
+// GetGVKFromDefinition help get Group Version Kind from DefinitionReference
+func GetGVKFromDefinition(dm discoverymapper.DiscoveryMapper, definitionRef v1alpha2.DefinitionReference) (schema.GroupVersionKind, error) {
+	var gvk schema.GroupVersionKind
+	groupResource := schema.ParseGroupResource(definitionRef.Name)
+	gvr := schema.GroupVersionResource{Group: groupResource.Group, Resource: groupResource.Resource, Version: definitionRef.Version}
+	kinds, err := dm.KindsFor(gvr)
+	if err != nil {
+		return gvk, err
+	}
+	if len(kinds) < 1 {
+		return gvk, &meta.NoResourceMatchError{
+			PartialResource: gvr,
+		}
+	}
+	return kinds[0], nil
 }
 
 // Object2Unstructured convert an object to an unstructured struct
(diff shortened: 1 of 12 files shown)


pkg/oam/util/helper.go
```diff
@@ -257,6 +259,23 @@ func GetDefinitionName(dm discoverymapper.DiscoveryMapper, u *unstructured.Unstr
...
+	kinds, err := dm.KindsFor(gvr)
+	if err != nil {
+		return gvk, err
+	}
+	if len(kinds) < 1 {
+		return gvk, &meta.NoResourceMatchError{
+			PartialResource: gvr,
+		}
+	}
+	return kinds[0], nil
```
ryanzhang-oss, line 276: I am not sure that this is a good assumption. How can there be multiple kinds when the version is set?

wonderflow: This is used for cases whose version was not set.

pkg/oam/util/helper.go
```diff
@@ -297,6 +299,23 @@ func GetDefinitionName(dm discoverymapper.DiscoveryMapper, u *unstructured.Unstr
 	return mapping.Resource.Resource + "." + groupVersion.Group, nil
 }
 
+// GetGVKFromDefinition help get Group Version Kind from DefinitionReference
+func GetGVKFromDefinition(dm discoverymapper.DiscoveryMapper, definitionRef v1alpha2.DefinitionReference) (schema.GroupVersionKind, error) {
```
hongchaodeng, line 303: I don't see any place that uses this method.

hongchaodeng: Could you provide an example how this is gonna be used?

hongchaodeng (approved):
code lgtm

Please provide use cases for GVKFromDefinition method

pkg/oam/util/helper.go
wonderflow: sorry, I added one in the code.

Issue #270, opened 2020-10-26
[BUG]Two examples of oam: flight-tracker and dependency do not authorize oam's permissions to crd, so the example will not run.


linjiemiao:
When I deploy flight-tracker, oam does not have permission to ingress, so it will report an error:
```
[...]

PR #271, opened 2020-10-27
[ISSUE #270] Bug fixed, OAM example can not run correctlly.

linjiemiao:
**I fixed four examples in total:**

1） component revsion name format changed in example **component-versioning**
2)   add clusterrole [KIND=Foo] in dependency example
3)   add clusterrole [KIND=Ingress] in flight-tracker example
4)  change example-health-scope's workloadrefs in typed-component example

### DIFF
--- a/examples/flight-tracker/tracker-app-config.yaml
+++ b/examples/flight-tracker/tracker-app-config.yaml
@@ -7,7 +7,20 @@ data:
   endpoint: ZGItd29ya2xvYWQ=
   password: ZGJwYXNzd29yZA==
   username: cG9zdGdyZXM=
-
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: flight-tracker-clusterrole
+  labels:
+    rbac.oam.dev/aggregate-to-controller: "true"
+rules:
+  - apiGroups:
+      - extensions
+    resources:
+      - 'ingresses'
+    verbs:
+      - '*'
 ---
 apiVersion: core.oam.dev/v1alpha2
 kind: ApplicationConfiguration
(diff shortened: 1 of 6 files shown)


examples/flight-tracker/tracker-app-config.yaml
```diff
@@ -79,6 +92,11 @@ spec:
...
-                  servicePort: 8080
\ No newline at end of file
+              rules:
+                - host: x.oam.dev
+                  http:
+                    paths:
+                      - backend:
+                          serviceName: web-ui
+                          servicePort: 8080
+                        path: /
```
wonderflow, line 102: Users don't own this domain and could not view it if we setting like this. Please keep the original way, thanks.

examples/typed-component/sample_application_config.yaml
```diff
@@ -8,10 +8,7 @@ spec:
   workloadRefs:
     - apiVersion: core.oam.dev/v1alpha2
       kind: ContainerizedWorkload
-      name: example-component
-    - apiVersion: core.oam.dev/v1alpha2
-      kind: ContainerizedWorkload
-      name: example-component2
```
wonderflow, line 14: why remove

linjiemiao: Because Application does not contain a component named example-component2 or example-component1。It contains a component named web-service-component.

examples/flight-tracker/tracker-app-config.yaml
linjiemiao: ok, i will keep the original way.

examples/typed-component/sample_application_config.yaml
wonderflow: I see, thanks

wonderflow (approved):
Excellent Job!

PR #278, opened 2020-11-03
fix panic for missing latestRevision while render component

captainroy-hy:
It's possible to raise a panic while render revision-enabled Component.
The problem is found in [e2e-test for component-versioning](https://github.com/crossplane/oam-kubernetes-runtime/blob/4ea4b295981d929b7a5b3285aab770333fb943c6/test/e2e-test/component_version_test.go#L307). 

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/render.go
@@ -56,6 +56,7 @@ const (
 	errFmtSetParam         = "cannot set parameter %q"
 	errFmtUnsupportedParam = "unsupported parameter %q"
 	errFmtRequiredParam    = "required parameter %q not specified"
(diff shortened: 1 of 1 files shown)


pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -56,6 +56,7 @@ const (
 	errFmtSetParam         = "cannot set parameter %q"
 	errFmtUnsupportedParam = "unsupported parameter %q"
 	errFmtRequiredParam    = "required parameter %q not specified"
+	errFmtCompRevision     = "cannot get latest revision for component %q"
```
wonderflow, line 59:
```suggestion
	errFmtCompRevision     = "cannot get latest revision for component %q while revision is enabled"
```

wonderflow (approved):

PR #283, opened 2020-11-06
use update instead of patch on applying trait

### LINKED ISSUE #272: Trait apply should use update instead of patch
## Problem Statement

Currently, when oam-runtime applies traits, it does MergePatch (code [link1](https://github.com/crossplane/crossplane-runtime/blob/378a57303ea0a2d13ae21668044364ad8d1035d9/pkg/resource/api.go#L149) and [link2](https://github.com/kubernetes-sigs/controller-runtime/blob/74fd294a89c65c8efc17ab92e0d2014d36e357a8/pkg/client/patch.go#L100)). Basically, it compares the trait to apply (A), and existing trait object in cache (B), and compare difference between A and B to create a patch to send to APIServer.

This will lead to the following problem:
1. Trait initialized as:
  ```
  auto: false
  rules: [a, b, c]
  ```
2. Trait updated as:
  ```
  auto: true
  ```

In above scenario, we expect `rules` field to be removed, but when we apply it, the trait actually becomes:
```
auto: true
ruels: [a, b, c]
```

## Proposed Solution

To solve the problem, we should just update the spec of trait instead of patch [here](https://github.com/crossplane/oam-kubernetes-runtime/blob/696eb7b021d03865ccdaa74bf5effb207a410ce6/pkg/controller/v1alpha2/applicationconfiguration/apply.go#L98).
For fields safety concerns, the traits are operational data so it will only be managed by OAM platform. Therefore updating is OK in this case. On the contrary, workloads could be changed by end users and we should consider more about how to update it.

captainroy-hy:
to fix #272 

Solution: Add a new `updatingClient` to update trait while retain `patchingClient` to apply workload.

add e2e-test


pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -75,17 +76,25 @@ func (fn WorkloadApplyFns) Finalize(ctx context.Context, ac *v1alpha2.Applicatio
 }
 
 type workloads struct {
-	client    resource.Applicator
-	rawClient client.Client
-	dm        discoverymapper.DiscoveryMapper
+	patchingClient resource.Applicator
+	updatingClient resource.Applicator
```
wonderflow, line 80: I think we should answer why we need two client with same type here. Everyone review this code will have the same question. Add the answer to annotation here.

test/e2e-test/trait_update_test.go
```diff
@@ -0,0 +1,196 @@
+package controllers_test
```
wonderflow, line 1: please try refactor it to be unit-test, thanks! Only necessary cases use e2e test, this could be a unit test like dependency

pkg/controller/v1alpha2/applicationconfiguration/apply.go
captainroy-hy: Yep. 
- patchingClient will be used to patching-apply Workload objs while updatingClient will be used to updating-apply Trait objs.  As #272 described, patching-apply cannot remove fields that are present before applying but absent now, that's not expected for Trait. So we use update instead of patch for applying Trait.
- The applying strategy on Workload is TBD, so just keep patchingClient unchanged.

test/e2e-test/trait_update_test.go
captainroy-hy: got it.

pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -95,7 +98,7 @@ func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus,
 				continue
 			}
 			t := trait.Object
-			if err := a.client.Apply(ctx, &trait.Object, ao...); err != nil {
+			if err := a.updatingClient.Apply(ctx, &trait.Object, ao...); err != nil {
```
hongchaodeng, line 101: Why not change Apply() to update? That's more clear. Otherwise it still looks similar to `kubectl apply`.

captainroy-hy: The primary concern is that if use client.Update(), we need more code, e.g., Get/Check-IfNotFound-Create/IfFound-SetResourceVersion-Update, just like what underlying of Apply() has done. So maybe we can just re-use Apply() here.

wonderflow (approved):
Good Job!

pkg/controller/v1alpha2/applicationconfiguration/apply.go
hongchaodeng: Sounds good!

PR #291, opened 2020-11-10
fix unstable unit tests of AppConfig controller

captainroy-hy:
In unit test, after applying Creat/Update/Apply of AppConfig or other related resources, it's necessary to check whether applying is successful before triggering the reconciliation of AppConfig. Otherwise, it's possible to fail the reconciliation for the inconsistent resources.


hongchaodeng (approved):
lgtm

PR #295, opened 2020-11-11
enable HealthScope to check version-enabled component

### LINKED ISSUE #288: [Feature] scope should include all workload instance with versioning mechanism
Currently, oam scope only have one target workload for one component. But with versioning mechanism, one component could have multiple workload instances, we should support this case.

captainroy-hy:
to fix #288 

HealthScope will check all existing workload instances referring to different versions of one component.
If ANY workload is unhealthy, the overall condition is unhealthy.
For those who have multiple workloads, the health condition returned will contain all diagnosis or workload status of all.
And format as below screenshot:
image

BTW, as for the number of healthy/total workloads shown in summary, multiple workload instances belonging to one component still only count 1.

- add unit tests

### DIFF
--- a/pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
+++ b/pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
@@ -39,10 +39,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


wonderflow (approved):

wonderflow:
GOOD JOB!

PR #297, opened 2020-11-13
Refine health diagnosis output 

captainroy-hy:
- Refine the output of HealthScope when target workload has peer versioning workloads.
The diagnosis will sort by the revision number in descending order.
And if the revision order is larger than 3 and its required replicas is 0,  don't show its diagnosis info.
```yaml
status:
  healthConditions:
  - componentName: example-component
    diagnosis: 'example-component-v6:Ready:1/1  example-component-v5:Ready:1/1  example-component-v4:Ready:0/0 '
    healthStatus: HEALTHY
    targetWorkload:
      kind: ContainerizedWorkload
      name: example-component-v6
      uid: 4242f59b-0cb9-4cd9-8641-fd128d59cd5a
```
- change HealthScope work interval from 1min => 10sec
- fix unstable unit test

### DIFF
--- a/pkg/controller/v1alpha2/applicationconfiguration/dependency_test.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/dependency_test.go
@@ -86,7 +86,6 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 
 	// common function for verification
 	verify := func(appConfigName, reason string) {
-
 		appconfigKey := client.ObjectKey{
 			Name:      appConfigName,
 			Namespace: namespace,
@@ -104,7 +103,7 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 
 		By("Reconcile")
 		req := reconcile.Request{NamespacedName: appconfigKey}
-		Expect(func() error { _, err := reconciler.Reconcile(req); return err }()).Should(BeNil())
+		reconcileRetry(reconciler, req)
 
 		outFooKey := client.ObjectKey{
 			Name:      outName,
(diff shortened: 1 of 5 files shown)


pkg/controller/v1alpha2/applicationconfiguration/dependency_test.go
```diff
@@ -517,17 +504,13 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
...
+		reconcileRetry(reconciler, req)
 
 		By("Verify the appconfig's dependency is satisfied")
 		Eventually(
 			func() []v1alpha2.UnstaifiedDependency {
 				tempAppConfig := &v1alpha2.ApplicationConfiguration{}
-				err := k8sClient.Get(ctx, appconfigKey, tempAppConfig)
-				if err != nil || tempAppConfig.Status.Dependency.Unsatisfied != nil {
-					// Try 3 (= 1s/300ms) times
-					reconciler.Reconcile(req)
```
wonderflow, line 529: why not keep the reconcile here？

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -394,3 +369,65 @@ func getVersioningPeerWorkloadRefs(ctx context.Context, c client.Reader, wlRef r
...
+		for _, peerHC := range peerHCs[1:] {
+			basic.WorkloadStatus = fmt.Sprintf("%s %s:%s",
+				basic.WorkloadStatus,
+				peerHC.TargetWorkload.Name,
+				peerHC.WorkloadStatus)
+		}
+	} else {
+		basic.Diagnosis = fmt.Sprintf("%s:%s", peerHCs[0].TargetWorkload.Name, peerHCs[0].Diagnosis)
+		for i, peerHC := range peerHCs[1:] {
+			if i > 1 && peerHC.Diagnosis == fmt.Sprintf(infoFmtReady, 0, 0) {
```
wonderflow, line 423:
```suggestion
			if i > 0 && peerHC.Diagnosis == fmt.Sprintf(infoFmtReady, 0, 0) {
```

I hope we just keep one

pkg/controller/v1alpha2/applicationconfiguration/dependency_test.go
captainroy-hy: Got it, I will revert the change here. It's diff from other reconcile logic.

pkg/controller/v1alpha2/core/scopes/healthscope/healthscope.go
captainroy-hy: Sure.

wonderflow (approved):
LGTM，thanks

PR #298, opened 2020-11-16
add ApplyOnceOnly switch for production running requirement

### LINKED ISSUE #277: [Feature] support don't reconcile if AppConfig or Component not updated
**Is your feature request related to a problem? Please describe.**

Currently, OAM runtime will reconcile periodically, this mechanism can help CR always working as expected ( unexpected changes will be reconciled back periodically)

But it will introduce some risks especially when the oam-runtime Controller was just upgraded.

**Describe the solution you'd like**

I hope we can add a switch on oam-runtime, by default the switch is turned off, it will work as it is now. If the switch is turned on, the controller will check AppConfig generation, if generation is the same with status.observedGeneration and the condition is ready, it will not reconcile.

captainroy-hy:
to fix #277 
Add a "ApplyOnceOnly(bool type)" switch for AppConfig controller.  By default it's false.
Once it's enabled, the reconciler will not affect underlying workload and trait CR if no generation changed.
As for Create & Delete events occurring on AppConfig and Component, it works as before.
But for Update events, reconciliation is triggered when the object's metadata.labels/annotations/finalizers, or spec is changed.

### DIFF
--- a/cmd/oam-kubernetes-runtime/main.go
+++ b/cmd/oam-kubernetes-runtime/main.go
@@ -52,6 +52,8 @@ func main() {
 	flag.BoolVar(&logCompress, "log-compress", true, "Enable compression on the rotated logs.")
 	flag.IntVar(&controllerArgs.RevisionLimit, "revision-limit", 50,
 		"RevisionLimit is the maximum number of revisions that will be maintained. The default value is 50.")
+	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
+		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
 	flag.Parse()
 
 	// setup logging
@@ -100,6 +102,9 @@ func main() {
 		os.Exit(1)
 	}
 	oamLog.Info("starting the controller manager")
+	if controllerArgs.ApplyOnceOnly {
+		oamLog.Info("applyOnceOnly is enabled that means workload or trait only apply once if no spec change even they are changed by others")
+	}
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		oamLog.Error(err, "problem running manager")
 		os.Exit(1)
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -18,4 +18,8 @@ type Args struct {
 	// RevisionLimit is the maximum number of revisions that will be maintained.
 	// The default value is 50.
 	RevisionLimit int
+
+	// ApplyOnceOnly indicates whether workloads and traits should be
+	// affected if no spec change is made in the ApplicationConfiguration.
+	ApplyOnceOnly bool
 }
--- a/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
+++ b/pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -97,21 +98,23 @@ func Setup(mgr ctrl.Manager, args controller.Args, l logging.Logger) error {
 		}).
 		Complete(NewReconciler(mgr, dm,
 			WithLogger(l.WithValues("controller", name)),
-			WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+			WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
+			WithApplyOnceOnly(args.ApplyOnceOnly)))
 }
 
 // An OAMApplicationReconciler reconciles OAM ApplicationConfigurations by rendering and
 // instantiating their Components and Traits.
 type OAMApplicationReconciler struct {
-	client     client.Client
-	components ComponentRenderer
-	workloads  WorkloadApplicator
-	gc         GarbageCollector
-	scheme     *runtime.Scheme
-	log        logging.Logger
-	record     event.Recorder
-	preHooks   map[string]ControllerHooks
-	postHooks  map[string]ControllerHooks
+	client        client.Client
+	components    ComponentRenderer
+	workloads     WorkloadApplicator
+	gc            GarbageCollector
+	scheme        *runtime.Scheme
+	log           logging.Logger
+	record        event.Recorder
+	preHooks      map[string]ControllerHooks
+	postHooks     map[string]ControllerHooks
+	applyOnceOnly bool
 }
 
 // A ReconcilerOption configures a Reconciler.
@@ -165,6 +168,14 @@ func WithPrehook(name string, hook ControllerHooks) ReconcilerOption {
 func WithPosthook(name string, hook ControllerHooks) ReconcilerOption {
 	return func(r *OAMApplicationReconciler) {
 		r.postHooks[name] = hook
+	}
+}
+
+// WithApplyOnceOnly indicates whether workloads and traits should be
+// affected if no spec change is made in the ApplicationConfiguration.
+func WithApplyOnceOnly(applyOnceOnly bool) ReconcilerOption {
+	return func(r *OAMApplicationReconciler) {
+		r.applyOnceOnly = applyOnceOnly
 	}
 }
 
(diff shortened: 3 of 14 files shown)


pkg/controller/controller.go
```diff
@@ -18,4 +18,9 @@ type Args struct {
 	// RevisionLimit is the maximum number of revisions that will be maintained.
 	// The default value is 50.
 	RevisionLimit int
+
+	// LimitedReconcile indicates whether reconcile periodically or only reconcile
+	// when specific fields are changed, e.g., metadata.labels, metadata.annotations,
+	// and spec.
+	LimitedReconcile bool
```
wonderflow, line 25: I suggest we name it `ReconcileByChangeOnly`

cmd/oam-kubernetes-runtime/main.go
```diff
@@ -52,6 +52,8 @@ func main() {
 	flag.BoolVar(&logCompress, "log-compress", true, "Enable compression on the rotated logs.")
 	flag.IntVar(&controllerArgs.RevisionLimit, "revision-limit", 50,
 		"RevisionLimit is the maximum number of revisions that will be maintained. The default value is 50.")
+	flag.BoolVar(&controllerArgs.LimitedReconcile, "limited-reconcile", false,
```
wonderflow, line 55:
```suggestion
	flag.BoolVar(&controllerArgs.LimitedReconcile, "reconcile-change-only", false,
```

cmd/oam-kubernetes-runtime/main.go
```diff
@@ -52,6 +52,8 @@ func main() {
...
+		"Enable only reconcile when crucial update event occurs, meanwhile disable reconcile periodically")
```
wonderflow, line 56:
```suggestion
		"If enabled, controller will only reconcile when object spec is changed for successfully reconciled object. If not, the controller will periodically reconcile. By default, it's disabled")
```

pkg/controller/controller.go
captainroy-hy: Got it.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -313,14 +330,17 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
...
+		return reconcile.Result{RequeueAfter: dependCheckWait}, nil
 	}
 
 	// the posthook function will do the final status update
-	return reconcile.Result{RequeueAfter: waitTime}, nil
+	if r.limitedReconcile {
+		return reconcile.Result{}, nil
+	}
+	// enforce to reconcile periodically, even no event occurs
+	return reconcile.Result{RequeueAfter: longWait}, nil
```
wonderflow, line 343: We should not implement like this.

We should check if the spec is unchanged and already reconciled succeed.

I assume at least we should have these conditions:

1. metadata.generation==status.observedGeneration.
2. the component spec(revision) is the same with the living workload revision.
3. the status is reconciled succeed
4. the workload is health
5. the trait CR is working normally (I assume we can check at least 1) condition is ready 2) generation=status.observedGeneration)

wonderflow (commented):
You could write an e2e test for this feature as it's not easy to implement as unit-test.

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -279,7 +290,12 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 	log.Debug("Successfully rendered components", "workloads", len(workloads))
 	r.record.Event(ac, event.Normal(reasonRenderComponents, "Successfully rendered components", "workloads", strconv.Itoa(len(workloads))))
 
-	if err := r.workloads.Apply(ctx, ac.Status.Workloads, workloads, resource.MustBeControllableBy(ac.GetUID())); err != nil {
+	applyOpts := []resource.ApplyOption{resource.MustBeControllableBy(ac.GetUID())}
+	if r.applyOnceOnly {
+		log.Debug("ApplyOnceOnly is enabled")
```
wonderflow, line 295:
```suggestion
		log.Info("ApplyOnceOnly is enabled")
```

This message deserves info level

wonderflow: please move this log at start time, not for all CR reconcile

pkg/controller/v1alpha2/applicationconfiguration/render.go
```diff
@@ -145,6 +145,11 @@ func (r *components) renderComponent(ctx context.Context, acc v1alpha2.Applicati
 	}
 	util.AddLabels(w, compInfoLabels)
 
+	compInfoAnnotations := map[string]string{
+		oam.AnnotationAppGeneration: fmt.Sprintf("%d", ac.Generation),
```
wonderflow, line 149: strconv.Itoa is better

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -544,3 +560,40 @@ func eligible(namespace string, ws []v1alpha2.WorkloadStatus, w []Workload) []un
 
 	return eligible
 }
+
+// ApplyOnceOnlyError indicates the resource being applied has no spec change
+// comparing to the existing one.
+type ApplyOnceOnlyError struct{}
```
wonderflow, line 566:
```suggestion
type GenerationUnchanged struct{}
```

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -544,3 +560,40 @@ func eligible(namespace string, ws []v1alpha2.WorkloadStatus, w []Workload) []un
...
+
+func (e *ApplyOnceOnlyError) Error() string {
+	return "skip applying since no change occurs on this resource"
```
wonderflow, line 569:
```suggestion
	return "apply-only-once enabled, and detect generation in the annotation unchanged, will not apply"
```

pkg/controller/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -544,3 +560,40 @@ func eligible(namespace string, ws []v1alpha2.WorkloadStatus, w []Workload) []un
...
+
+func applyOnceOnly() resource.ApplyOption {
+	return func(ctx context.Context, current, desired runtime.Object) error {
+		c, _ := current.(metav1.Object)
+		d, _ := desired.(metav1.Object)
+		if c == nil || d == nil {
+			return errors.Errorf("invalid object being applied: %q ",
+				desired.GetObjectKind().GroupVersionKind())
+		}
+		cLabels, dLabels := c.GetLabels(), d.GetLabels()
```
wonderflow, line 580: what will happen if one of them is nil? For example, it's not update, it's create the first time

captainroy-hy: For Create operation, it will not come into this logic.

pkg/controller/v1alpha2/applicationconfiguration/update_trait_test.go
```diff
@@ -226,6 +227,8 @@ spec:
         - trait:
             apiVersion: example.com/v1
             kind: Bar
+            matadata:
+              name: example-trait
```
wonderflow, line 231: why we need a name here?

wonderflow (commented):
The code LGTM, please hold until we test in real production case. I will merge it after test good.

pkg/controller/v1alpha2/applicationconfiguration/update_trait_test.go
captainroy-hy: If no fixed trait name, I have to get the trait name from AppConfig status (need more code) or use the implicitly generated trait name by the controller.

captainroy-hy:
> The code LGTM, please hold until we test in real production case. I will merge it after test good.

yep, got it.

cmd/oam-kubernetes-runtime/main.go
```diff
@@ -100,6 +102,9 @@ func main() {
 		os.Exit(1)
 	}
 	oamLog.Info("starting the controller manager")
+	if controllerArgs.ApplyOnceOnly {
+		oamLog.Info("applyOnceOnly is enabled")
```
wonderflow, line 106:
```suggestion
		oamLog.Info("applyOnceOnly is enabled that means workload or trait only apply once if no spec change even they are changed by others")
```

wonderflow (approved):
LGTM， excellent job!

PR #299, opened 2020-11-16
Fix sample code and add READMEs

dsyer:
The typed-component sample didn't work at all. And both that 
and the containerized-workload sample (as mentioned in top
level README) lacked any documentation to indicate how to
use them and verify that they were working correctly.

### DIFF
--- a/examples/typed-component/README.md
+++ b/examples/typed-component/README.md
@@ -1,0 +1,56 @@
+# Custom Workload
+
+This is an example web application with a custom workload
+
+## Run ApplicationConfiguration
+
+Install the components.
+
+```bash
+$ kubectl apply -f .
+workloaddefinition.core.oam.dev/web-service created
+scopedefinition.core.oam.dev/healthscopes.core.oam.dev created
+healthscope.core.oam.dev/example-health-scope created
+applicationconfiguration.core.oam.dev/example-appconfig created
+component.core.oam.dev/web-service-component created
+```
+
+
+> NOTE: The `oam-k8s-runtime` webhook is needed to enhance the workload in the `Component` definition. If you are running without the webhook you will need to add the following to the `spec.workload` in `sample_component.yaml`:  
+>
+>      apiVersion: core.oam.dev/v1alpha2
+>      kind: ContainerizedWorkload
+
+
+## Result
+
+A `web-service-component` is created (running Wordpress, which you will see on the corresponding `Service` endpoint):
+
+```
+$ kubectl get all
+NAME                                         READY   STATUS    RESTARTS   AGE
+pod/web-service-component-78fbdd6787-5nwh5   1/1     Running   0          2m19s
+pod/web-service-component-78fbdd6787-9gmfp   1/1     Running   0          2m20s
+pod/web-service-component-78fbdd6787-r652l   1/1     Running   0          2m21s
+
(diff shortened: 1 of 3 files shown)


dsyer:
I'm not sure the `typed-component` is working as designed even with this change. It seems to make no difference if I delete the `workload.spec.type`. So I guess it's actually not doing anythin?

examples/typed-component/sample_component.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 spec:
   workload:
     type: web-service
+    apiVersion: core.oam.dev/v1alpha2
```
resouer, line 8: `type` is mutually exclusive with `apiVersion` and `kind`, so please leave this yaml as is. Typed component is a experimental design in oam-runtime and it's not compliance with OAM spec v0.2.1.

Also, we decide not to accept this format in OAM spec per recent community meeting, so `type` will be deprecated soon.

examples/typed-component/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+NAME                                    READY   UP-TO-DATE   AVAILABLE   AGE
+deployment.apps/web-service-component   3/3     3            3           2m22s
+
+NAME                                               DESIRED   CURRENT   READY   AGE
+replicaset.apps/web-service-component-6b4b8b57b7   0         0         0       2m22s
+replicaset.apps/web-service-component-78fbdd6787   3         3         3       2m21s
+```
+
+Wordpress will respond on port 80. There is a UI, but since there is no database it will not be fully functional.
```
resouer, line 41: Would you mind to change the image to sth else like `nginx`? Wordpress doesn't make sense to me unless a database component is included.

resouer (commented):
Thank you Deve, some comments left.

examples/typed-component/sample_component.yaml
dsyer: If type is deprecated we should remove it though? That’s the opposite. So don’t know what to do.

examples/typed-component/README.md
dsyer: It’s not my sample. I just used what was there.

examples/typed-component/sample_component.yaml
resouer: Yes, exactly, it should be removed. That's why I suggest leave it as is unless you want to take this burden (I don't think so, and it's safer to do clean up in separate PR).

examples/typed-component/README.md
resouer: LGTM then, you've already done your part!

resouer:
LGTM when `typed-component` section is restored. A follow up PR will be send to clean up the samples. /cc @dsyer

resouer (approved):
LGTM after typed-component section is restored. Maintainers will send follow up PR to clean up the samples. /cc @dsyer

examples/typed-component/sample_component.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 spec:
   workload:
     type: web-service
```
wonderflow, line 7: the typed component can work if we enable webhook of oam-k8s-runtime

dsyer:
Restored component definition and put a note about it in the README. That should resolve all the comments above.

wonderflow (approved):
Thanks! @dsyer

PR #310, opened 2020-11-22
implement trait.ApplyTo through AppConfig validating webhook

### LINKED ISSUE #14: Honor the AppliesTo field in the trait
The current AC controller implementation doesn't look up to a trait's TraitDefinition CR when it applies a trait to a workload. We need to throw an error if the workload is not part of the 'appliesTo' field.

### LINKED ISSUE #50: AppliesTo should support API Group
For example:

```console
$ kubectl oam traits
NAME   DEFINITION                       APPLIES TO
route     route.core.oam.dev           containerizedworkload,apps.k8s.io
```

This means the trait route can be applied to any workloads under apps.k8s.io group.

/cc @ryanzhang-oss @wonderflow

captainroy-hy:
fix #50 & fix #14 
- implement trait.ApplyTo through AppConfig validating webhook
trait.ApplyTo can specify appliable workload in 3 formats: workloadType, workload DefinitionRef name, API Group.  
e.g., ApplyTo: []string{`autoscaler`, `manuscalers.core.oam.dev`,`standard.oam.dev`}
- refactor appConfig validating webhook 
to avoid repetitive GET/unmarshal operations in multiple validation funcs, use a new `ValidatingAppConfig` typed struct to prepare all data before validation
- fix unit tests

### DIFF
--- a/pkg/webhook/v1alpha2/applicationconfiguration/helper.go
+++ b/pkg/webhook/v1alpha2/applicationconfiguration/helper.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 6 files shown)


pkg/webhook/v1alpha2/applicationconfiguration/helper_test.go
```diff
@@ -1,116 +1,16 @@
...
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
-	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam/mock"
 )
 
-func TestCheckComponentVersionEnabled(t *testing.T) {
```
wonderflow, line 21: why delete this test?

captainroy-hy: It's moved to [here](func TestValidateWorkloadNameForVersioningFn(t *testing.T) {) and refactored.

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -16,32 +15,102 @@ import (
...
+		if err := json.Unmarshal(comp.Spec.Workload.Raw, &wl); err != nil {
+			allErrors = append(allErrors, errors.Wrapf(err,
+				errFmtUnmarshalWorkload, tmp.compName))
+		}
+		tmp.workloadContent = wl
+
+		// get workload definition
+		wlDef, err := util.FetchWorkloadDefinition(ctx, c, dm, &wl)
+		if err != nil {
+			allErrors = append(allErrors, errors.Wrapf(err,
```
wonderflow, line 85: we don't return all errors at once, please directly return this error.

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
```diff
@@ -16,32 +15,102 @@ import (
...
+				allErrors = append(allErrors, errors.Wrapf(err,
+					errFmtGetTraitDefinition, tmp.compName))
+			}
+			tmpT.traitDefinition = *tDef
+			tmp.validatingTraits = append(tmp.validatingTraits, tmpT)
 		}
 	}
-	return false, nil
+	// remove nil error
+	return allErrors
```
wonderflow, line 113: just return nil, if no errors, allErrors cause confusion

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -23,25 +21,40 @@ import (
...
 
 	// WorkloadNamePath indicates field path of workload name
 	WorkloadNamePath = "metadata.name"
 )
 
 var appConfigResource = v1alpha2.SchemeGroupVersion.WithResource("applicationconfigurations")
 
+// AppConfigValidator provides functions to validate ApplicationConfiguration
+type AppConfigValidator interface {
+	Validate(context.Context, client.Client, discoverymapper.DiscoveryMapper, ValidatingAppConfig) []error
```
wonderflow, line 38: I see most cases only `ValidatingAppConfig ` is used, why don't make it more clean like:
```suggestion
	Validate(context.Context, ValidatingAppConfig) []error
```

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -23,25 +21,40 @@ import (
...
+
 // ValidatingHandler handles CloneSet
 type ValidatingHandler struct {
 	Client client.Client
 	Mapper discoverymapper.DiscoveryMapper
 
 	// Decoder decodes objects
 	Decoder *admission.Decoder
+
+	DefaultValidators []AppConfigValidator
```
wonderflow, line 57:
```suggestion
	Validators []AppConfigValidator
```

I don't see all of them are defaults, if so, who're not default ones?

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -67,97 +80,117 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
+	var allErrs []error
+	for _, c := range v.validatingComps {
+		workloadType := c.component.GetLabels()[oam.WorkloadTypeLabel]
+		workloadDefRefName := c.workloadDefinition.Spec.Reference.Name
+		// TODO version of workloadDefRef should be taken into account
+		workloadGroup := c.workloadDefinition.GetObjectKind().GroupVersionKind().Group
+		for _, t := range c.validatingTraits {
+			if len(t.traitDefinition.Spec.AppliesToWorkloads) == 0 {
+				// AppliesToWorkloads is empty, the trait can be applied to ANY workload
+				continue
```
wonderflow, line 174: Both empty and "*" cases.

Don't forget `*`

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -67,97 +80,117 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
+			}
+			appliable := false
+			for _, applyTo := range t.traitDefinition.Spec.AppliesToWorkloads {
+				if workloadType == applyTo ||
+					workloadDefRefName == applyTo ||
+					workloadGroup == applyTo {
```
wonderflow, line 180: if the case is group, it should be like `*.<group>`, don't miss the `*.` prefix

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -67,97 +80,117 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
-		if len(workloadName) != 0 {
-			return false, fmt.Sprintf(reasonFmtWorkloadNameNotEmpty, workloadName)
+// ValidateTraitAppliableToWorkloadFn validates whether a trait is allowed to apply to the workload.
+func ValidateTraitAppliableToWorkloadFn(_ context.Context, _ client.Client, _ discoverymapper.DiscoveryMapper, v ValidatingAppConfig) []error {
+	klog.Info("validate applicationConfiguration", "name", v.appConfig.Name)
+	var allErrs []error
+	for _, c := range v.validatingComps {
+		workloadType := c.component.GetLabels()[oam.WorkloadTypeLabel]
+		workloadDefRefName := c.workloadDefinition.Spec.Reference.Name
+		// TODO version of workloadDefRef should be taken into account
```
wonderflow, line 169: what does it mean?

captainroy-hy: Since [definition is allowed to have more than one version](https://github.com/crossplane/oam-kubernetes-runtime/blob/2cf8f6353b3c40eecdbe915bf199ec0544d848b9/apis/core/v1alpha2/core_types.go#L34), maybe ApplyTo is possible to indicate a specific version of WorkloadDefinition in the future.

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
captainroy-hy: Maybe ValidateFunc needs to use client&dm in its own logic in the future, even though current ValidateFuncs don't need because prepared data is sufficient for them.

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
captainroy-hy: got it.

pkg/webhook/v1alpha2/applicationconfiguration/helper.go
captainroy-hy: got it.

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -67,97 +81,117 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
-		c, _, err := util.GetComponent(ctx, client, acc, appConfig.GetNamespace())
-		if err != nil {
-			return false, fmt.Sprintf(errFmtCheckWorkloadName, err.Error())
+// ValidateWorkloadNameForVersioningFn validates workload name for version-enabled component
+func ValidateWorkloadNameForVersioningFn(_ context.Context, v ValidatingAppConfig) []error {
+	var allErrs []error
+	for _, c := range v.validatingComps {
+		for _, t := range c.validatingTraits {
+			if t.traitDefinition.Spec.RevisionEnabled {
+				goto Validate
```
wonderflow, line 145: using `break ` is enough in this case

pkg/webhook/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -67,97 +81,117 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
...
-		if len(workloadName) != 0 {
-			return false, fmt.Sprintf(reasonFmtWorkloadNameNotEmpty, workloadName)
+// ValidateTraitAppliableToWorkloadFn validates whether a trait is allowed to apply to the workload.
+func ValidateTraitAppliableToWorkloadFn(_ context.Context, v ValidatingAppConfig) []error {
+	klog.Info("validate applicationConfiguration", "name", v.appConfig.Name)
+	var allErrs []error
+	for _, c := range v.validatingComps {
+		workloadType := c.component.GetLabels()[oam.WorkloadTypeLabel]
+		workloadDefRefName := c.workloadDefinition.Spec.Reference.Name
+		// TODO version of workloadDefRef should be taken into account
```
wonderflow, line 167:
```suggestion
		// TODO consider a CRD group could have multiple version and maybe we need to specify the minimum version here in the future
```

wonderflow (approved):
LGTM after nit and rebase

wonderflow (approved):

PR #317, opened 2020-12-28
add data passing mechanism in oam-kubernetes-runtime

hoopoe61:
There are data passing requirements among collaborative applications. Such as: 
1. When Application A needs to call the service provided by database D, D will pass the address of database to A.
2. When a request needed to be processed by more than one services, this request needed to be passed from the first application to the second one, and so on.
The resource dependency in OAM has implemented condition judgement in dependency but with no support for data transfer. The Service Binding only supports passing static data among objects without complicate data processing.

This proposal aims to realizing the data passing mechanism in OAM based on Resource Dependency in OAM, and specifically include:
1. The data passing among objects during the creation phase. This mechanism can judge the dependencies of different objects in one application configuration, and pass dynamic data between objects.
2. The data processing capability. During data passing, one object can collect many data and then process as needed, such as jsonPath.
The following example shows how to pass data from one component to another one.

```
metadata:
  name: sample-application-config
spec:
  components:
    - componentName: sample-output-component
      dataOutputs:
      - name: sample-dataoutput
        fieldPath: "status.key"
        # outputToRef is used to describe how to get data from sample-output-component and write to an intermediate storage
        outputToRef:
          apiVersion: v1
          kind: ConfigMap
          name: sample-config
          operations:
          - type: "jsonPatch"
            op: "add"
            toFieldPath: "data.redis"
            value: '"{}"'
            conditions:
            - op: "notEq"
              value: ""
              fieldPath: "status.instanceID"
          - type: "jsonPatch"
            op: "add"
            toFieldPath: "data.redis"
            toPath: "instanceID"
            valueFrom:
              fieldPath: "status.instanceID"
            conditions:
            - op: "notEq"
              value: ""
              fieldPath: "status.instanceID"
        conditions:
        - op: "notEq"
          value: ""
          fieldPath: "status.instanceID"
    - componentName: sample-input-component
      dataInputs:
      - valueFrom:
          dataOutputName: "sample-dataoutput"
        toFilePaths: []string{"spec.key"}
        # valueFromRef is used to describe how to get data from the intermediate storage and write to sample-input-component
          apiVersion: v1
          kind: ConfigMap
          name: sample-config
          operations:
          - type: "jsonPatch"
            op: "add"
            toFieldPath: "spec.containers.0.env.0.value"
            valueFrom:
              fieldPath: "data.redis"
            conditions:
            - op: "notEq"
              value: ""
              fieldPath: "data.redis"
        conditions:
        - op: "notEq"
          value: ""
          fieldPath: "status.key"
```


apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
...
+}
+
+// DataOperation defines the specific operation for data
+type DataOperation struct {
+	// TypeName specific the type of DataOperation
+	TypeName string `json:"type"`
+	// Operator specific the operation under this DataOperation type
+	Operator DataOperator `json:"op"`
+	// FieldPath refers to the value of an object's field.
+	ToFieldPath string `json:"toFieldPath"`
```
hongchaodeng, line 515: Just `fieldPath`

apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
...
+	// Path refers to the value of an object's specfied by FieldPath.
+	ToPath string `json:"toPath,omitempty"`
```
hongchaodeng, line 517: just `path`

apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
...
+	// +optional
+	// Value specifies an expected value
+	// This is mutually exclusive with ValueFrom
+	Value string `json:"value,omitempty"`
+	// +optional
+	// ValueFrom specifies expected value from AppConfig
+	// This is mutually exclusive with Value
+	ValueFrom  ValueFrom              `json:"valueFrom,omitempty"`
```
hongchaodeng, line 525: Why add `ValueFrom` to dataOutput?

apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
...
+	Conditions []ConditionRequirement `json:"conditions,omitempty"`
```
hongchaodeng, line 526: Why add `Conditions` to dataOutput?

apis/core/v1alpha2/core_types.go
```diff
@@ -504,6 +546,11 @@ type DataInput struct {
 
 	// ToFieldPaths specifies the field paths of an object to fill passed value.
 	ToFieldPaths []string `json:"toFieldPaths,omitempty"`
+
+	Conditions []ConditionRequirement `json:"conditions,omitempty"`
```
hongchaodeng, line 550: Why add `Conditions` to dataInput?

apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
...
+type DataReference struct {
+	runtimev1alpha1.TypedReference `json:",inline"`
+	// Operations specify the data processing operations
+	Operations []DataOperation `json:"operations,omitempty"`
+}
+
+// DataOperation defines the specific operation for data
+type DataOperation struct {
+	// TypeName specific the type of DataOperation
+	TypeName string `json:"type"`
```
ryanzhang-oss, line 511: What are the possible "type"s?

apis/core/v1alpha2/core_types.go
ryanzhang-oss, line 517: What's the difference between these two fields? It's pretty confusing.

apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
...
+	// Operator specific the operation under this DataOperation type
+	Operator DataOperator `json:"op"`
+	// FieldPath refers to the value of an object's field.
+	ToFieldPath string `json:"toFieldPath"`
+	// Path refers to the value of an object's specfied by FieldPath.
+	ToPath string `json:"toPath,omitempty"`
+	// +optional
+	// Value specifies an expected value
+	// This is mutually exclusive with ValueFrom
+	Value string `json:"value,omitempty"`
```
ryanzhang-oss, line 521: Why is there another Value here? I saw it can be just ""

apis/core/v1alpha2/core_types.go
ryanzhang-oss: Co-ask

ryanzhang-oss (changes requested):
We probably need a better example or a design doc for this feature. It's not very intuitive to understand the usage of this feature from the existing example and descriptions.

hongchaodeng:
ref https://github.com/oam-dev/kubevela/issues/770

apis/core/v1alpha2/core_types.go
```diff
@@ -494,8 +494,50 @@ type DataOutput struct {
 	// If no conditions is specified, it is by default to check output value not empty.
 	// +optional
 	Conditions []ConditionRequirement `json:"conditions,omitempty"`
+	// OutputToReference specifies the object used to store of intermediate data genereted by Operations
+	OutputToReference DataReference `json:"outputToRef,omitempty"`
```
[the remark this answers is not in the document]
hoopoe61: OutputToReference in dataoutput is corresponding with valueFromReference in datainput. In addition,  OutputToReference and valueFromReference is also used in yaml. StoreReference is better than DataReference

apis/core/v1alpha2/core_types.go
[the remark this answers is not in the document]
hoopoe61: good idea

apis/core/v1alpha2/core_types.go
hoopoe61: ToFieldPath is used in DataOperation to distinguish it from the FieldPath in ValueFrom. ToFieldPath means writing data to this filedPath.

apis/core/v1alpha2/core_types.go
hoopoe61: same as ToFieldPath

apis/core/v1alpha2/core_types.go
hoopoe61: The valueFrom in DataOutput is used to pick up data in specified path of an object in every DataOperation. For example, the status.InstanceID and  status.InstanceStatus of a kafka instance are picked up and written to configmap. The FieldPath in former data dependency implementation is also reserved for compatibility.

apis/core/v1alpha2/core_types.go
hoopoe61: The conditions in dataOutput are used to specify the conditions needed for data output. For example, the info of a kafka instance can be used only when its creation is completed. So the condition that status.InstacneStatus is not empty must be set.

apis/core/v1alpha2/core_types.go
hoopoe61: This is similar with conditions in dataOutput. These conditions are used to represent the data in intermediate storage is ready now and can be passed to other objects.

apis/core/v1alpha2/core_types.go
hoopoe61: In now implementation, only jsonPath is supported.

apis/core/v1alpha2/core_types.go
hoopoe61: For example, we want to pass the info of redis with configmap. As the redis has many attributes, so a json is use as the value. The following format may be used:
apiVersion: v1
data:
  instanceID: r-8vba95ca8ba75c74
  redis: '{"instanceClass":"redis.***","instanceID":"***","status":"***"}'
kind: ConfigMap

apis/core/v1alpha2/core_types.go
hoopoe61: The Value may be used to set some field. For example, when there is no "redis" key in the data of a configmap, we will use the following data operations to initialize the redis:
operations:
          - type: "jsonPatch"
            op: "add"
            toFieldPath: "data.redis"
            value: '"{}"'

apis/core/v1alpha2/core_types.go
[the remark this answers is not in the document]
hongchaodeng: OutputToReference -> OutputStore

apis/core/v1alpha2/core_types.go
hongchaodeng: how about change to `ToDataPath`?

apis/core/v1alpha2/core_types.go
hongchaodeng: Need to split DataReference into two structs -- OutputStore and InputStore. They have different fields.

hongchaodeng (commented):
After offline discussion, put some comments.

Additionally, need to increase coverage to pass CI.

hongchaodeng (approved):
lgtm

@ryanzhang-oss Has some questions. Please see if they are addressed. Or ping @hoopoe61  offline if needed.

wonderflow:
Hi, would you like to send this PR to [kubevela](https://github.com/oam-dev/kubevela)

hoopoe61:
> @hoopoe61 Hi, would you like to send this PR to [kubevela](https://github.com/oam-dev/kubevela)

    Yes, I will send this PR to kubevela next week.

hongchaodeng:
@ryanzhang-oss 
You have placed a request here. Is it resolved? Could you review again? Ping @hoopoe61 offline if needed.

ryanzhang-oss (approved):
it will be good that the data input/output logic is in its own module like how dependency works.

PR #323, opened 2021-03-18
feat: add debug-flag for enabling logs

### LINKED ISSUE #284: [Feature] make development log setting as an argument
image

Currently the development flag is always true, so we will get lot's of debug log by default.

I think we should make the default development flag to be false, and add a command in Makefile such as `make debug-run` to switch the flag on.

kushthedude:
Signed-off-by: kushthedude <kushthedude@gmail.com>

fixes #284

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,10 @@ run: go.build
 	@# To see other arguments that can be provided, run the command with --help instead
 	$(GO_OUT_DIR)/$(PROJECT_NAME)
 
+# This is for running as a local process for convenience with debug logs.
+debug-run:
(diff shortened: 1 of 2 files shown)


kushthedude:
Hi team, I am not sure if lint-failure is due to the changes in the PR?

resouer:
@kushthedude Seems not, it should be OK after re-run.

wonderflow (approved):
Good job!

wonderflow:
@kushthedude Would you like to fix the same issue in KubeVela https://github.com/oam-dev/kubevela/blob/master/cmd/core/main.go#L120

kushthedude:
Sur @wonderflow !

Issue #327, opened 2021-04-20
[BUG] delete component and modify appConfig's components to  null array lead to panic


xishengcai:
**Describe the bug**

**To Reproduce**
1. deploy oam latest version
2. kubectl apply -f examples/containerized-workload/
3. kubectl delete component example-component
4. modify applicationconfiguration:   example-appconfig. spec.components: []

**Expected behavior**
expect deployment deleted

**Screenshots**
![image](https://user-images.githubusercontent.com/21197150/115354365-216bd180-a1ec-11eb-840d-ee16be1a2ec7.png)

**Cluster information**

- Kubernetes Version : 1.17

PR #329, opened 2021-04-20
fix contorller: check the number of workloads before applying workloads

yangsoon:
fix https://github.com/crossplane/oam-kubernetes-runtime/issues/327

check the number of workloads before applying workloads, 
fix the problem that the array is out of bounds.


pkg/controller/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -100,6 +100,9 @@ type workloads struct {
 }
 
 func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
+	if len(w) == 0 {
```
ryanzhang-oss, line 103: please add reason for this change and add tests

wonderflow (approved):

